PR #7, opened 2021-02-25
Add set/get support for objects

### LINKED ISSUE #2: Support for objects?
Feature request:
I see references to things like `convertObjCObjectToJSIValue` and the reverse but mmkv.mm doesn't seem to be exposing this functionality.

Would it be possible to get native object/dict support so one could avoid using `JSON.stringify()` (and rely on the native counter-part)? Or have you found that it's more trouble than it's worth?

mrousavy:
Adds support for:

```ts
const user = {
  username: 'Marc',
  age: 20
}

MMKV.set(user, 'user')
```

and

```ts
const user = MMKV.get('user') // --> { username: 'Marc', age: 20 }
```

This is very WIP and I'm not sure if it even works since I think MMKV (protobuf) has to know the size of the object it should retrieve.

Closes #2

### DIFF
--- a/ios/Mmkv.mm
+++ b/ios/Mmkv.mm
@@ -37,6 +37,9 @@ static void install(jsi::Runtime & jsiRuntime)
         } else if (arguments[0].isString()) {
             auto stringValue = convertJSIStringToNSString(runtime, arguments[0].getString(runtime));
             [MMKV.defaultMMKV setString:stringValue forKey:keyName];
+        } else if (arguments[0].isObject()) {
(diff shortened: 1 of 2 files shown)


mrousavy:
I think on android I have to store bytes manually 😧

Screenshot 2021-02-25 at 10 33 43

scarlac:
Dang ok. Sorry, I was hoping that the code was secretly ready on both platforms but figured there might be a good reason why it wasn't already integrated. Encoding the bytes essentially means reimplementing json.stringify again which is probably a bit much.

Issue #11, opened 2021-02-28
Web


nandorojo:
Could be cool to add web support, maybe with store or just localStorage directly.

mrousavy:
Hi! Unfortunately I am not really familiar with react-native-web, and have no use-case for it.

It looks like this is relatively easy to implement though, probably only requires a bit of fiddling around with configuration files and then just creating a `index.web.ts`.

I would however gladly accept any PR regarding web support 👍

mrousavy:
Working on this here: https://github.com/mrousavy/react-native-mmkv/pull/130

mrousavy:
Would love some help with testing 🙏

nandorojo:
@mrousavy I think this will break in an app with SSR, since `window` is undefined.

I would make `storage` a getter function instead, which is only called when one of the functions is called.

mrousavy:
does that not work?
https://github.com/mrousavy/react-native-mmkv/blob/5b4d0ad07c561a8b69b241f1d8d712dcfa0cee14/src/createMMKV.web.ts#L7

nandorojo:
No, because `window` is accessed when it's inexistent.

A null check is insufficient.

A typical solution on SSR apps is to do this:

`if (typeof window !== 'undefined')`

Since that won't throw an error.

In your case, if you changed that line to 

```ts
const storage = () => global.localStorage ?? window.localStorage ?? localStorage;
```

Then it would work fine. Since it's now a function, we only access `window` when one of the MMKV functions is called, and no one would call it on the server.

To be safe, you could run the window check I used above.

However, a more standardized RNW check to see if we can access the dom is `canUseDOM`. See the usage here of that variable here:

https://gist.github.com/nandorojo/066ff2f40419b7e06054cc7282e24f8d#file-hoverable-ts-L11

nandorojo:
The issue is, you're throwing an error if storage doesn't exist, but it will never exist on the server

nandorojo:
I added the suggested edits here: https://github.com/mrousavy/react-native-mmkv/pull/130#issuecomment-913179225

mrousavy:
done!

vishwanathkokare:
localStorage → It's synchronous, slow, and limited, which can block the UI

mrousavy:
> It's synchronous

Good, this is what MMKV is about.

> slow

Do you have any benchmarks? What is a faster synchronous storage library?

> limited

How? It supports every feature of MMKV, except for encryption.

PR #45, opened 2021-04-01
Add Web Support

PhysicsAreBad:
Addressing suggestion in #11 with localStorage support.


mrousavy:
Hi! Thanks for the PR! I've never used RN Web before - how do I test this out? Could you also add a web app to the example/ so I can test it?

src/index.web.ts
```diff
@@ -0,0 +1,72 @@
...
+  set: (key: string, value: boolean | string | number): void => {
+    localStorage.setItem(key, value.toString());
+  },
+  /**
+   * Get a boolean value for the given `key`.
+   *
+   * @default false
+   */
+  getBoolean: (key: string): boolean => {
+    var result = localStorage.getItem(key);
```
mrousavy, line 14: Please use `let` instead of `var`.

src/index.web.ts
```diff
@@ -0,0 +1,72 @@
...
+   * @default undefined
+   */
+  getString: (key: string): string | undefined => {
+    var result = localStorage.getItem(key);
+
+    if (result) {
+      return result;
+    }
+
+    return undefined;
```
mrousavy, line 34: I feel like this can simply be 
```js
return result;
```

src/index.web.ts
```diff
@@ -0,0 +1,72 @@
...
+   * @default []
+   */
+  getAllKeys: (): string[] => {
+    var keys = [] as string[];
+
+    for (var i = 0; i < localStorage.length; i++) {
+      keys.push(localStorage.key(i) as string);
+    }
+
+    return keys;
```
mrousavy, line 68: Can't we just use
```js
return Object.keys(localStorage);
```
?

mrousavy (changes requested):
Left some comments, once those are fixed and I have a react-native-web example in the `example/` folder we can merge this 👍

src/index.web.ts
PhysicsAreBad: Unfortunately, since local storage returns null instead of undefined if there is no result, I have to do to this to maintain the same functionality.

LinusU: How about:

```suggestion
    return result ?? undefined;
```

zhigang1992: Actually the check above should also have been changed to 

```typescript
if (result != null) {
```

Otherwise empty string will fall into the undefined / null path.

PhysicsAreBad:
Made suggested changes @mrousavy

src/index.web.ts
```diff
@@ -0,0 +1,54 @@
...
+  },
+  /**
+   * Get a string value for the given `key`.
+   *
+   * @default undefined
+   */
+  getString: (key: string): string | undefined => {
+    let result = localStorage.getItem(key);
+
+    return result != null ? result : undefined;
```
LinusU, line 26: In TypeScript (and modern JavaScript) this can be written as:

```suggestion
    return result ?? undefined;
```

PhysicsAreBad: Thanks, I did not realize that `??` does not follow standard JS truthy and instead only null/undefined.

example/src/Benchmarks.ts
```diff
@@ -14,6 +14,7 @@ export const benchmarkAgainstAsyncStorage = async () => {
     const start = global.performance.now();
     for (let i = 0; i < iters; i++) {
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      // @ts-ignore
```
mrousavy, line 17: why?

example/src/Benchmarks.ts
```diff
@@ -23,6 +24,7 @@ export const benchmarkAgainstAsyncStorage = async () => {
...
```
mrousavy, line 27: why?

example/tsconfig.json
```diff
@@ -0,0 +1,62 @@
+{
```
mrousavy, line 1: I believe I have configured the example to use the tsconfig from the root folder, is there a specific reason you added a second one specifically for the example?

example/tslint.json
```diff
@@ -0,0 +1,34 @@
+{
```
mrousavy, line 1: same as with the tsconfig - example should use eslintrc from root folder

mrousavy (changes requested):
okay I have left a few nitpicking comments, I'd like them to be resolved (only affects example). After that, I think everything looks perfect 👌

example/src/Benchmarks.ts
PhysicsAreBad: Originally, there was an issue with webpack not working since unused variables were not allowed in the original typescript set-up. This has been resolved with changing the main tsconfig to allow unused variables.

PhysicsAreBad: Originally, there was an issue with webpack not working since unused variables were not allowed in the original typescript set-up. This has been resolved with changing the main tsconfig to allow unused variables.

PhysicsAreBad:
I made the further suggested edits.

mrousavy:
So I ran the App for iOS and Android and nothing seems to be broken there, so that's good. However, when I run `yarn web` I get the following error:

```
example 🔥 yarn web
yarn run v1.22.10
$ cd web && webpack serve
ℹ ｢wds｣: Project is running at http://localhost:8080/
ℹ ｢wds｣: webpack output is served from /
ℹ ｢wds｣: Content not from webpack is served from /Users/mrousavy/Projects/react-native-mmkv-pr/example/web
✖ ｢wdm｣: asset vendor.js 1.33 MiB [emitted] (name: vendor) 1 related asset
asset app.js 387 KiB [emitted] (name: app) 1 related asset
asset index.html 392 bytes [emitted]
runtime modules 51.4 KiB 20 modules
cacheable modules 1.28 MiB
  modules by path ../node_modules/webpack-dev-server/ 21.1 KiB 11 modules
  modules by path ../node_modules/react-dom/ 901 KiB 6 modules
  modules by path ../node_modules/html-entities/lib/*.js 61 KiB 5 modules
  modules by path ../node_modules/querystring/*.js 4.51 KiB 3 modules
  modules by path ../node_modules/react/ 70.6 KiB 2 modules
  modules by path ../node_modules/webpack/hot/*.js 1.42 KiB
    ../node_modules/webpack/hot/emitter.js 75 bytes [built] [code generated]
    ../node_modules/webpack/hot/log.js 1.34 KiB [built] [code generated]
  modules by path ../node_modules/url/*.js 23.1 KiB
    ../node_modules/url/url.js 22.8 KiB [built] [code generated]
    ../node_modules/url/util.js 314 bytes [built] [code generated]
../node_modules/webpack/hot/ sync nonrecursive ^\.\/log$ 170 bytes [built] [code generated]

ERROR in app
Module not found: Error: path argument is not a string

ERROR in app
Module not found: Error: path argument is not a string

webpack 5.30.0 compiled with 2 errors in 4681 ms
ℹ ｢wdm｣: Failed to compile.
```

@PhysicsAreBad

mrousavy:
Actually the native app also breaks by this PR. This must be some issue with the new babel config:


```
yarn run v1.22.10
$ react-native start --reset-cache
                                                      
                        #######                       
                   ################                   
                #########     #########               
            #########             ##########          
        #########        ######        #########      
       ##########################################     
      #####      #####################       #####    
      #####          ##############          #####    
      #####    ###       ######       ###    #####    
      #####    #######            #######    #####    
      #####    ###########    ###########    #####    
      #####    ##########################    #####    
      #####    ##########################    #####    
      #####      ######################     ######    
       ######        #############        #######     
         #########        ####       #########        
              #########          #########            
                  ######### #########                 
                       #########                      
                                                      
                                                      
warning: the transform cache was reset.
                    Welcome to Metro!
              Fast - Scalable - Integrated


transform[stderr]: Could not resolve "/Users/mrousavy/Projects/react-native-mmkv-pr/src/index" in file /Users/mrousavy/Projects/react-native-mmkv-pr/example/src/App.tsx.
transform[stderr]: Could not resolve "/Users/mrousavy/Projects/react-native-mmkv-pr/src/index" in file /Users/mrousavy/Projects/react-native-mmkv-pr/example/src/Benchmarks.ts.
 BUNDLE  ./index.tsx 

error: node_modules/react-native/Libraries/Components/DatePicker/DatePickerIOS.ios.js: /Users/mrousavy/Projects/react-native-mmkv-pr/example/node_modules/react-native/Libraries/Components/DatePicker/DatePickerIOS.ios.js: Missing class properties transform.
  114 |  */
  115 | class DatePickerIOS extends React.Component<Props> {
> 116 |   static DefaultProps: {|mode: $TEMPORARY$string<'datetime'>|} = {
      |   ^
  117 |     mode: 'datetime',
  118 |   };
  119 |
 BUNDLE  ./index.tsx 

error: node_modules/react-native/Libraries/Components/MaskedView/MaskedViewIOS.ios.js: /Users/mrousavy/Projects/react-native-mmkv-pr/example/node_modules/react-native/Libraries/Components/MaskedView/MaskedViewIOS.ios.js: Missing class properties transform.
  64 |  */
  65 | class MaskedViewIOS extends React.Component<Props> {
> 66 |   _hasWarnedInvalidRenderMask = false;
     |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  67 |
  68 |   render(): React.Node {
  69 |     const {maskElement, children, ...otherViewProps} = this.props;

```

A quick stackoverflow search [showed me a possible approach](https://stackoverflow.com/questions/35517245/error-missing-class-properties-transform/49553497#49553497), however I don't think that is a _solution_ but rather a wacky workaround. Can we make the babel config only apply to the web project and leave the normal config for the native apps as is? Surely react-native-web has a solution for this.

package.json
```diff
@@ -112,6 +113,9 @@
           "useTabs": false
         }
       ]
+    },
+    "globals": {
+      "localStorage": true
```
LinusU, line 118: Potentially better to add `/* global localStorage */` to the top of `src/index.web.ts`, since it's only available as a global on the web platform, and not to the rest of the code...

ref: https://eslint.org/docs/rules/no-undef

package.json
```diff
@@ -53,6 +53,7 @@
     "@react-native-community/eslint-config": "^2.0.0",
     "@release-it/conventional-changelog": "^2.0.0",
     "@types/jest": "^26.0.0",
+    "@types/node": "^14.14.37",
```
LinusU, line 56: Sorry if I'm missing something obvious, but I cannot find where these types are used?

Types in the `@types/node` package are only available when running inside Node.js, not in the browser which this new code is targeting, right?

nandorojo:
I think it would be easier to have a separate example folder for web, and just use a minimal expo web template.

mrousavy:
RN MMKV now supports web 🎉

PR #48, opened 2021-04-09
added ios minimum deployment target to md doc

myckhel:


mrousavy:
Hi! Thanks for the PR.

Doesn't cocoapods warn you that the library you're trying to install has a higher minimum iOS version than your Podfile?

mrousavy:
Also, if you are on a version lower than iOS 11, could you try if this library works anyways? Just edit `node_modules/react-native-mmkv/react-native-mmkv.podspec` and change the iOS version to something lower (like 9) then run `pod install` again. I wonder why I set that to 11 in the first place 🤔

myckhel:
since it is compatible with IOS 9 and instead of modifying the `node_modules/react-native-mmkv/react-native-mmkv.podspec`  would it be a good idea to make a new pull request downgrading the IOS to 9.0?

mrousavy:
If you can confirm that it works without problems and you can store and read values from the database on iOS 9, yes - we can set the version to iOS 9 in the podspec. Can you confirm that it works?

myckhel:
Oooooops!

cant confirm if it works on IOS 9.

this packages requires react-native@63.4 which support minimum IOS 10.0.

so i guess the minimum IOS should be 10.0 at least.

Jalson1982:
@mrousavy any plan to release a new version with a minimum deployment target of 10?

mrousavy:
@Jalson1982 can you patch your podspec locally as I described in the comment above to see if it actually works on iOS 10? If you can confirm it works, create a PR to change it to version 10 and I'll release a new version :v:

Jalson1982:
Yeah sure i will do it today :)

Jalson1982:
So I made a test:

Screenshot 2021-04-13 at 14 08 47

All works fine :)

myckhel:
Nice!


PR should be the next thing for you to do.

myckhel:
i will close this PR since new PR will be submitted for the case.

PR #49, opened 2021-04-09
added package namespace to android's ExampeJSIPackage code

myckhel:


INSTALL.md
```diff
@@ -53,8 +53,11 @@ To install MMKV with Reanimated, open your Android project (the `android` folder
 2. Right click, "New" > "Java class"
 3. Call it whatever you prefer, in my case it's `ExampleJSIPackage` because my app is called "Example".
 4. Add the following code:
+5. Replace `com.example` with your package namespace
```
mrousavy, line 56: 1. let's move that down (below the code snippet), because in 4. I have a `:` at the end.
2. if we add it here, we should also add it to the other steps (the one without reanimated)

mrousavy (changes requested):

mrousavy (approved):
looks good, thanks! 👍

omprakashpmfby (commented):
perfecrt

PR #96, opened 2021-07-08
Add clearAll

janicduplessis:
Adds the clear all method to delete all keys. This avoid having to loop through all keys and call delete.


mrousavy:
Thank you Janic! 🖤🚀

mrousavy:
released in [v1.2.0](https://github.com/mrousavy/react-native-mmkv/releases/tag/v1.2.0)

PR #130, opened 2021-09-05
feat: Web support

mrousavy:
Adds support for MMKV on the Web.

This uses the [`localStorage` API](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) instead of MMKV as a "native" instance.

On _native_ (iOS/Android) nothing changes.

Solves #11

### DIFF
--- a/src/createMMKV.web.ts
+++ b/src/createMMKV.web.ts
@@ -1,17 +1,18 @@
 import type { MMKVConfiguration, MMKVInterface } from 'react-native-mmkv';
 
-// @ts-expect-error global func is a native JSI func
 export const createMMKV = (config: MMKVConfiguration): MMKVInterface => {
-  // TODO: Implement custom ID
+  const prefix = config.id;
   // TODO: Implement Encryption
   // TODO: Implement custom Path?
   return {
     clearAll: () => localStorage.clear(),
-    delete: (key) => localStorage.removeItem(key),
-    set: (key, value) => localStorage.setItem(key, value),
-    getString: (key) => localStorage.getItem(key),
-    getNumber: (key) => Number(localStorage.getItem(key)),
-    getBoolean: (key) => Boolean(localStorage.getItem(key)),
-    getAllKeys: () => Object.keys(localStorage),
+    delete: (key) => localStorage.removeItem(`${prefix}.${key}`),
+    set: (key, value) => localStorage.setItem(`${prefix}.${key}`, value),
+    getString: (key) => localStorage.getItem(`${prefix}.${key}`) ?? undefined,
+    getNumber: (key) => Number(localStorage.getItem(`${prefix}.${key}`) ?? 0),
+    getBoolean: (key) =>
+      Boolean(localStorage.getItem(`${prefix}.${key}`) ?? false),
+    getAllKeys: () =>
+      Object.keys(localStorage).map((key) => key.replace(`${prefix}.`, '')),
   };
 };
--- a/src/createMMKV.web.ts
+++ b/src/createMMKV.web.ts
@@ -1,26 +1,35 @@
 /* global localStorage */
 import { Platform } from 'react-native';
 import type { MMKVConfiguration, MMKVInterface } from 'react-native-mmkv';
+import { canUseDOM } from 'fbjs/lib/ExecutionEnvironment';
 
 export const createMMKV = (config: MMKVConfiguration): MMKVInterface => {
-  // @ts-expect-error
-  const storage = global.localStorage ?? window.localStorage ?? localStorage;
-  if (storage == null) {
-    throw new Error(
-      `Could not find 'localStorage' instance! Platform: ${Platform.OS}`
-    );
-  }
-  const prefix = config.id;
+  const storage = () => {
+    if (!canUseDOM) {
+      throw new Error(
+        'Tried to access storage on the server. Did you forget to call this in useEffect?'
+      );
+    }
+    const domStorage =
+      global?.localStorage ?? window?.localStorage ?? localStorage;
+    if (!domStorage) {
+      throw new Error(
+        `Could not find 'localStorage' instance! Platform: ${Platform.OS}`
+      );
+    }
+    return domStorage;
+  };
+
+  // TODO: Support custom instances?
   // TODO: Implement Encryption
   // TODO: Implement custom Path?
   return {
-    clearAll: () => storage.clear(),
-    delete: (key) => storage.removeItem(`${prefix}.${key}`),
-    set: (key, value) => storage.setItem(`${prefix}.${key}`, value),
-    getString: (key) => storage.getItem(`${prefix}.${key}`) ?? undefined,
-    getNumber: (key) => Number(storage.getItem(`${prefix}.${key}`) ?? 0),
-    getBoolean: (key) => Boolean(storage.getItem(`${prefix}.${key}`) ?? false),
-    getAllKeys: () =>
-      Object.keys(storage).map((key) => key.substring(prefix.length + 1)),
+    clearAll: () => storage().clear(),
+    delete: (key) => storage().removeItem(key),
+    set: (key, value) => storage().setItem(key, value),
+    getString: (key) => storage().getItem(key) ?? undefined,
+    getNumber: (key) => Number(storage().getItem(key) ?? 0),
+    getBoolean: (key) => Boolean(storage().getItem(key) ?? false),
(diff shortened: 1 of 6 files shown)


nandorojo:
Based on the conversation at #11, I re-wrote some of the logic to support SSR. Sorry the formatting is kinda off, my prettier is being weird.

```tsx
/* global localStorage */
import { Platform } from 'react-native';
import type { MMKVConfiguration, MMKVInterface } from 'react-native-mmkv';
import { canUseDOM } from 'fbjs/lib/ExecutionEnvironment'

export const createMMKV = (config: MMKVConfiguration): MMKVInterface => {
  const storage = () => {
    if (!canUseDOM) {
        throw new Error('[mmkv] Tried to access storage on the server. Did you forget to call this in useEffect?')
    }
    const domStorage = global.localStorage ?? window.localStorage ?? localStorage;
    if (!domStorage) {
        throw new Error(
            `[mmkv] Could not find 'localStorage' instance! Platform: ${Platform.OS}`
        );
    }
    return domStorage
  }

  const prefix = config.id;
  // TODO: Implement Encryption
  // TODO: Implement custom Path?
  return {
    clearAll: () => storage().clear(),
    delete: (key) => storage().removeItem(`${prefix}.${key}`),
    set: (key, value) => storage().setItem(`${prefix}.${key}`, value),
    getString: (key) => storage().getItem(`${prefix}.${key}`) ?? undefined,
    getNumber: (key) => Number(storage().getItem(`${prefix}.${key}`) ?? 0),
    getBoolean: (key) => Boolean(storage().getItem(`${prefix}.${key}`) ?? false),
    getAllKeys: () =>
      Object.keys(storage()).map((key) => key.substring(prefix.length + 1)),
  };
};
```

mrousavy:
@nandorojo thanks!

mrousavy:
what do you think about those 3 todos?
https://github.com/mrousavy/react-native-mmkv/blob/2cdac72d8e69b4eca65503144bdba554ba6fcbdc/src/createMMKV.web.ts#L23-L25

I think separating by instance ID will be a bit complicated because we have to prefix the keys, add a separator wildcard, ignore them when getting all keys, etc...

nandorojo:
I've been thinking about the open TODOs.

One issue is, if you don't create a custom path, then people might store things incorrectly. You can envision someone doing this:

```ts
const userStore = new MMKV({ id: 'user' })
const artistStore = new MMKV({ id: 'artist' })

const artistName = artistStore.getString('name')
const userName = userStore.getString('name')
```

One solution is to _always_ have a custom path. Keep a static ID incrementer, and if someone doesn't provide an ID, make it fall back to `__MMKV-${staticId}`.

```ts
let staticId = 0
export const createMMKV = (config?: MMKVConfiguration): MMKVInterface => {
  const storage = () => {
    if (!canUseDOM) {
        throw new Error('[mmkv] Tried to access storage on the server. Did you forget to call this in useEffect?')
    }
    const domStorage = global.localStorage ?? window.localStorage ?? localStorage;
    if (!domStorage) {
        throw new Error(`[mmkv] Could not find 'localStorage' instance! Platform: ${Platform.OS}`);
    }
    return domStorage
  }

  const prefix = config?.id || `__MMKV-${staticId++}`;
  // TODO: Implement Encryption
  // TODO: Implement custom Path?
  return {
    clearAll: () => storage().clear(),
    delete: (key) => storage().removeItem(`${prefix}.${key}`),
    set: (key, value) => storage().setItem(`${prefix}.${key}`, value),
    getString: (key) => storage().getItem(`${prefix}.${key}`) ?? undefined,
    getNumber: (key) => Number(storage().getItem(`${prefix}.${key}`) ?? 0),
    getBoolean: (key) => Boolean(storage().getItem(`${prefix}.${key}`) ?? false),
    getAllKeys: () =>
      Object.keys(storage()).map((key) => key.substring(prefix.length + 1)),
  };
};
```

Then, we could change `getAllKeys` to:

```ts
const getAllKeys = () => {
  const keys = []
  for (const key of Object.keys(storage())) {
    if (key.startsWith(`${prefix}.`)) {
      keys.push(key.substring(prefix.length + 1))
    }
  }
  return keys
}
```

mrousavy:
Using an incremented counter won't work, because then there is no way to determine which storage instance you're going to get if not using an ID. It might contain your data, it might contain data from another storage instance.

The prefix thing _would_ work, but what if the user has the ID `test.` and another storage instance with `test.another.`? That breaks the getAllKeys algorithm

mrousavy:
@nandorojo do you have any idea on how to approach this? LocalStorage doesn't seem to support scoping and using multiple instances out of the box, and prefixing keys with instance IDs is _very_ error prone.

I see two solutions:
1. Find a way to scope localStorage (aka create multiple separate instances, just like MMKV)
2. Don't allow creating multiple instances on the Web (throw error when trying to pass `id` or `path`)

Also, encryption is a whole other topic.

nandorojo:
I don't know of a real way to create multiple instances...so #2 might be the move. 

That sounds interesting re: web assembly, but I honestly have no idea how WA works.

I think to ship this, it would be easiest to not allow IDs on web. 

Even without an ID, do you think we should add a prefix on local storage though, like `__mmkv-`, to prevent clashing with other storage names?

mrousavy:
Alright, how do I test now if it works? 😅

nandorojo:
I think it should say `if (config.id && config.id !== ...)`

Currently it'll break if you don't specifically pass the default id.

As for testing it, I can help in 2 weeks, really busy with some things at the moment! But using an expo project and running `expo start web` should be good enough.

mrousavy:
The default ID is _always_ passed, even when you pass null. 

https://github.com/mrousavy/react-native-mmkv/blob/90dd516e12ad96a9aa73c12339d420cd7a420ee0/src/MMKV.ts#L116

axeldelafosse:
Happy to help for testing too!

nandorojo:
@mrousavy it's not, that's the native file.

See: https://github.com/mrousavy/react-native-mmkv/pull/130/files#diff-5cdb22ef6507d7454778bc0139b59874be49154336bbce3ff5462984ff3d770dR7

nandorojo:
Screen Shot 2021-11-12 at 4 59 02 PM

src/createMMKV.web.ts
```diff
@@ -0,0 +1,42 @@
+/* global localStorage */
+import type { MMKVConfiguration, NativeMMKV } from 'react-native-mmkv';
+
+const canUseDOM =
+  typeof window !== 'undefined' && window.document?.createElement != null;
+
+export const createMMKV = (config: MMKVConfiguration): NativeMMKV => {
```
nandorojo, line 7: This should be:

```ts
export const createMMKV = ({ id = 'mmkv.default', ...config }: MMKVConfiguration = {}): NativeMMKV => {
  if (id !== 'mmkv.default') {
    throw new Error("MMKV: 'id' is not supported on Web!");
  }
```

nandorojo: Once that's fixed, I think this is good to go.

mrousavy:
@nandorojo the `config` is `{ id: 'mmkv.default }` _by default_, so when you use:

```ts
new MMKV() // pass nothing
```

it is `{ id: 'mmkv.default }`. On native, and on web. Unless I'm missing something here, I don't think we need your change?

nandorojo:
I don't think that's correct. The web file has no such default set: https://github.com/mrousavy/react-native-mmkv/pull/130/files/2f031e36fd940b0b3a34f05d206a32fc9f0b5724#diff-5cdb22ef6507d7454778bc0139b59874be49154336bbce3ff5462984ff3d770d

nandorojo:
oh wait. I didn't realize its `new MMKV()`. Wow haha sorry about that. I thought `createMMKV` was the consumer's usage.

nandorojo:
Yeah this is good to go then 🤦‍♂️

src/createMMKV.web.ts
nandorojo: Scratch this, didn't realize that `new MMKV()` wraps this.

a-eid:
@nandorojo @mrousavy could you please explain why this check exists ? 

```ts
 if (!canUseDOM) {
        throw new Error('[mmkv] Tried to access storage on the server. Did you forget to call this in useEffect?')
    }
```

mrousavy:
I think this is because you cannot use `localStorage` on a server. This is a client side feature. See SSR

PR #135, opened 2021-09-06
fix: make bootstrap script work on windows

yfunk:
This PR adds the same fix from https://github.com/callstack/react-native-builder-bob/commit/8bac4d4 I added over at `vision-camera-image-labeler` to make the `yarn` override script work on Windows.


mrousavy:
thanks @yfunk! 🖤

PR #150, opened 2021-09-15
Add jsi.cpp to the list of libraries used in CMakeLists.txt

kelvinpompey:
This fixes the issue I had building an expo bare app. I also tested it with the included React Native CLI app to make sure it didn't break it. This library inspired me to learn about JSI. Thank you! 😊

### DIFF
--- a/android/CMakeLists.txt
+++ b/android/CMakeLists.txt
@@ -14,6 +14,7 @@ include_directories(
 
 add_library(reactnativemmkv  # <-- Library name
         SHARED
+        "${NODE_MODULES_DIR}/react-native/ReactCommon/jsi/jsi/jsi.cpp"
         src/main/cpp/cpp-adapter.cpp
         src/main/cpp/MmkvHostObject.cpp
 )


mrousavy:
Hey! Thanks for the PR! Unfortunately this is not the right approach to fix this, because now JSI is compiled multiple times (causing duplicate C++ symbols, once for react-native and once for react-native-mmkv), which causes any error messages to be swallowed instead of being propagated upwards as catchable JS errors - aka this makes the entire app crash (SIGABRT) when a native C++ error occurs.


I am wondering why this is the only way for you to get it working though, does `libreactnativejni.so` not exist in `android/build/**`?

kelvinpompey:
I see. Thanks for the explanation. I don't see libreactnativejni.so in android/build/** in my expo app. I don't see it in the library's React Native CLI app either.

kelvinpompey:
@mrousavy I don't know if there's another solution to this. It seems that other JSI libraries are using it as well. I saw this in the react-native-reanimated CMakeLists.txt. 

if(${REACT_NATIVE_TARGET_VERSION} LESS 64)
    set (JSI_PATH_FOR_RN_LESS_64 ${NODE_MODULES_DIR}/react-native/ReactCommon/jsi/jsi/jsi.cpp)
endif()

kelvinpompey:
@mrousavy I was wrong. libreactnativejni.so is in android/app/build/intermediates/merged_native_libs/debug/out/lib

Duell10111:
@kelvinpompey Your workaround also fails on my Windows machine with a syntax error: 

```
CMake Error at F:\Eigene Daten\Dokumente\FrozenFaces\Revoir\CallMe\node_modules\react-native-mmkv\android\CMakeLists.txt:17 (add_library):
Syntax error in cmake code when parsing string

      F:\Eigene Daten\Dokumente\FrozenFaces\Revoir\CallMe\android/../node_modules/react-native/ReactCommon/jsi/jsi/jsi.cpp

    Invalid character escape '\E'.
```

mrousavy:
The code from reanimated only applies to RN 0.63 and below, are you on a RN version below 64?

kelvinpompey:
@mrousavy The app uses React Native 0.63.4. That's what the current expo SDK uses.

PR #163, opened 2021-09-24
feat: Hooks and Listeners

mrousavy:
## Listeners

```ts
const mmkv = new MMKV()

mmkv.addOnValueChangedListener((changedKey) => {
  const newValue = mmkv.getString(changedKey) 
  console.log(`"${changedKey}" new value: ${newValue}`)
})
```

## Hooks

```tsx
function App() {
  const [username, setUsername] = useMMKVString("user.name")
  const [age, setAge] = useMMKVNumber("user.age")
  const [isPremiumUser, setIsPremiumUser] = useMMKVBoolean("user.isPremium")
}
```

```tsx
function App() {
  const mmkv = useMMKV()
  // ...
  mmkv.set("user.name", "Marc")
}
```

```tsx
function App() {
  const globalStorage = useMMKV()
  const userStorage = useMMKV({ id: `${userId}.storage` })

  const [username, setUsername] = useMMKVString("user.name", userStorage)
}
```

### DIFF
--- a/src/MMKV.ts
+++ b/src/MMKV.ts
@@ -1,3 +1,7 @@
+interface Listener {
+  remove: () => void;
+}
+
 /**
  * Used for configuration of a single MMKV instance.
  */
@@ -74,6 +78,12 @@ export interface MMKVInterface {
    * Delete all keys.
    */
   clearAll: () => void;
+  /**
+   * Adds a value changed listener.
+   */
+  addOnValueChangedListener: (
+    onValueChanged: (key: string) => void
+  ) => Listener;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


andreialecu:
This is awesome.

I wonder: Would an `useMMKVObject<T>` which auto uses JSON stringify/parse make sense? It's been something I added in my projects when using Async Storage in the past.

andreialecu:
Another thing that often comes up (but it can be a subsequent PR). 

Sometimes a way to use a dynamic array of strings and get the stored values for all of them is necessary (like asyncstorage multi-get, but in hook form). Not sure what the best way of implementing it would be.

Perhaps you could allow the first parameter (key) to take in an array of keys?

mrousavy:
@andreialecu Added `useMMKVObject<T>`

mrousavy:
Regarding Arrays: That's a bit more complex, especially with mixed types. With the current solution you can either hardcode the amount of values you need or use `useMMKVObject` and pass an array.

```ts
const [users, setUsers] = useMMKVObject<User[]>("users")
```

samuel3105 (approved):

PR #164, opened 2021-09-25
fix: typescript for JSI global functions

akinncar:
## Overview
Declared global functions that come from JSI to remove type error and `@ts-expect-error`.

image


mrousavy:
Thank you for your PR! 🙏🖤

Issue #137, opened 2021-09-07
Cannot create MMKV instance when passing either "path" or "encryptionKey"


t4ngth00:
The error "Error: Exception in HostFunction: <unknown>" when passing either "path" or "encryptionKey" to the initialized object of MMKV class's constructor.

Reproducible example:
import { MMKV } from 'react-native-mmkv';

const storage = new MMKV({
  id: 'some-id',
  path: 'some-path',
  encryptionKey: 'some-encryption-key',
});

I am using
react-native: 0.64.2
react-native-mmkv: 1.3.1

mrousavy:
well what's the exception? you didn't show me an error message

t4ngth00:
I've posted the error at the beginning of the issue. Here you are

ERROR  Error: Exception in HostFunction: <unknown>, js engine: hermes

mrousavy:
there's no error message in there? is there a stack trace being printed?

t4ngth00:
```
Accessing hidden method Landroid/os/Trace;->asyncTraceEnd(JLjava/lang/String;I)V (greylist, reflection, allowed)
Loaded version=93.0.4577.62 minSdkVersion=29 isBundle=true multiprocess=true packageId=2
Successfully loaded native library
Flushed 6 samples from 6 histograms.
flipper: FlipperClient::addPlugin flipper-plugin-redux-debugger
Creating MMKV instance "nr.care.default"... (Path: , Encryption-Key: )
<MMKV_IO.cpp:81::loadFromFile> loading [nr.care.default] with 3462 actual size, file size 8192, InterProcess 0, meta info version:3
<MMKV_IO.cpp:86::loadFromFile> loading [nr.care.default] with crc 833080238 sequence 2 version 3
<MMKV_IO.cpp:130::loadFromFile> loaded [nr.care.default] with 4 key-values
Creating MMKV instance "nr.care.encrypted"... (Path: /tmp/, Encryption-Key: some-encryption-key)
<MemoryFile.cpp:220::mkPath> some-encryption-key : Read-only file system
Error: Exception in HostFunction: <unknown>, js engine: hermes
[ERROR][NCS    ] camxncsintfqsee.cpp:2208 FillSensorData() Invalid sensor index -1
read panel temperature:26819
Invariant Violation: Module AppRegistry is not a registered callable module (calling runApplication). A frequent cause of the error is that the application entry file path is incorrect. 
[...]

PR #176, opened 2021-10-05
Fix cryptkey and path param order for Android

nairboh:
These params were out of order causing this issue when specifying either the path or encryption key on Android: https://github.com/mrousavy/react-native-mmkv/issues/137


mrousavy:
oh wow, what a dumb mistake. Thank you for the PR, am on mobile right now so I can't test - does supplying "path" and "encryptionKey" actually work now?

PR #213, opened 2021-10-27
Makes `react-native-mmkv` available on Apple TV 📺

todorone:
## Test plan

With this change, it compiles and works on `tvOS` 🎉

### DIFF
--- a/react-native-mmkv.podspec
+++ b/react-native-mmkv.podspec
@@ -10,7 +10,7 @@ Pod::Spec.new do |s|
   s.license      = package["license"]
   s.authors      = package["author"]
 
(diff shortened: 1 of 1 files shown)


jose920405:
This isn't true. I tried and received this error

`
ERROR  [runtime not ready]: Error: Failed to create a new MMKV instance: React Native is not running on-device. MMKV can only be used when synchronous method invocations (JSI) are possible. If you are using a remote debugger (e.g. Chrome), switch to an on-device debugger (e.g. Flipper) instead., js engine: hermes
`

Probably some extra steps are needed, but there is no information for running on TV

PR #219, opened 2021-11-04
update readme

axeldelafosse:
Link to the Expo installation discussion

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,10 @@ cd ios && pod install
 
 To correctly initialize MMKV on Android, please follow the [Installation guide](./INSTALL.md).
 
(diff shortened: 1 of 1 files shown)


mrousavy:
Axel thanks for the PR, but wouldn't it be better to directly include a `withMMKV.ts` config plugin in this repository? That way the user doesn't have to manually download and add it, it's just a simple `expo install react-native-mmkv` command

axeldelafosse:
Oh yeah it would be much better! But this solution is very sketchy, it's only temporary. I don't know what to do. I might try to find some time to do this but I'd prefer working on a better solution instead.

PR #244, opened 2021-12-02
Fix bug in useMMKV

robinsoncol:
When we initially call useMMKV, `lastConfiguration` is initialized, causing `!isConfigurationEqual(lastConfiguration.current, configuration)` to be `false` and we don't save a new MMKV instance. The solution I propose should fix this issue.

### DIFF
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -32,7 +32,10 @@ export function useMMKV(
   const instance = useRef<MMKV>();
 
   const lastConfiguration = useRef<MMKVConfiguration>();
-  if (lastConfiguration.current == null || !isConfigurationEqual(lastConfiguration.current, configuration)) {
+  if (
+    lastConfiguration.current == null ||
+    !isConfigurationEqual(lastConfiguration.current, configuration)
+  ) {
     lastConfiguration.current = configuration;
     instance.current = new MMKV(configuration);
   }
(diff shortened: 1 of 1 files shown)


src/hooks.ts
```diff
@@ -31,8 +31,9 @@ export function useMMKV(
 ): React.RefObject<MMKV> {
   const instance = useRef<MMKV>();
 
-  const lastConfiguration = useRef<MMKVConfiguration>(configuration);
-  if (!isConfigurationEqual(lastConfiguration.current, configuration)) {
+  const lastConfiguration = useRef<MMKVConfiguration>();
+  if (!lastConfiguration.current || !isConfigurationEqual(lastConfiguration.current, configuration)) {
```
mrousavy, line 35:
```suggestion
  if (lastConfiguration.current == null || !isConfigurationEqual(lastConfiguration.current, configuration)) {
```

mrousavy (changes requested):
Thanks for the PR! I prefer to use strict null checks (`== null`), could you please update that? Otherwise good to go, can't believe I missed that 😅

src/hooks.ts
robinsoncol: Done and thanks!

mrousavy:
Thank you!

PR #252, opened 2021-12-10
fix: handle case when key is changing

osdnk:
We have a given case:

```
const address = useAddress()
const [value] = useMMKVBoolean('state-' + address)
```

Currently, this is not working

### DIFF
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -85,7 +85,7 @@ function createMMKVHook<
       });
       return () => listener.remove();
     }, [key, mmkv]);
-    
+
     useEffect(() => {
       setValue(getter(mmkv, key));
     }, [key, mmkv]);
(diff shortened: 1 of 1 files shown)


mrousavy:
Thanks for the PR Michal, can you format your code so ESLint is happy?

mrousavy:
This runs on the first mount as well, so we have an additional `.get` on every key 🤔

mrousavy:
Hey @osdnk, did you find some time to work on this again to avoid the initial `.get` for every first call?

mrousavy:
Hey Michal hope you're doing well - let me know if https://github.com/mrousavy/react-native-mmkv/commit/d3ccb655bb3c8fee07b730e01e69f29c675675e2 fixed the issue for you!

PR #255, opened 2021-12-11
Fix extract tasks being added to clean tasks

janicduplessis:
The extract tasks are being incorrectly added to clean tasks (`externalNativeBuild{Debug|Release}Clean`) which causes files to be missing when running `./gradlew clean assembleRelease`.


mrousavy:
Thank you Janic! 🙏

PR #281, opened 2022-01-08
feat: support source builds on android

janicduplessis:
When building RN from source, the prebuilt artifacts will be located somewhere else. This detects source build by looking for existence of ReactAndroid project and if it is the case it will get artifacts from the right folder. We also need to depend on ReactAndroid build tasks to make sure the artifacts are built first.


mrousavy:
Looks good Janic! Seems like CI is happy too, let's ship it! Thank you! 🖤

PR #306, opened 2022-01-27
feat: Use `undefined` as a default value (**BREAKING CHANGE**)

mrousavy:
## What

This PR uses the new API I have created in my PR (https://github.com/Tencent/MMKV/pull/828) to return `undefined` instead of `0.0` (`getNumber(..)`) or `false` (`getBoolean(..)`) virtually without any added runtime cost.

## Why

Default values of `0.0` or `false` might be misleading in some cases.

## Changes

* `getNumber` now returns `undefined` instead of `0.0` if it was not found in the storage
* `getBoolean` now returns `undefined` instead of `false` if it was not found in the storage


jbreuer95:
+1

PR #312, opened 2022-02-02
feat: Add `MMKV.defaultInstance`

mrousavy:
```ts
const storage = MMKV.defaultInstance
```


mrousavy:
Actually this PR is redundant.

An instance created with `new MMKV()` is already the default, shared instance.

Issue #179, opened 2021-10-08
Change `encryptionKey` on existing instances


franjs:
From the docs on [IOS](https://github.com/Tencent/MMKV/wiki/iOS_advance#encryption) and [Android](https://github.com/Tencent/MMKV/wiki/android_advance#encryption) I can see that it is possible to change the encryption key for MMVK. 

It looks that this feature has not been implemented yet. Are there any plans to include this feature?

mattijsf:
Yes this would be a very useful addition. We would like to use that in the exact way.

mrousavy:
Alright I got 40 minutes of free time, what do you want the function to be called? I'm thinking `storage.recrypt(...)` - not sure if that's a word tho

mattijsf:
Cool. Well... here they call it `reKey` not sure if that has a good ring to it either: https://github.com/Tencent/MMKV/blob/3b79dad46d09881aed7a6b54291069130e53e249/flutter/README.md#encryptio

Quoting there docs:
```
// an unencrypted MMKV instance
var mmkv = MMKV("test-encryption");

// change from unencrypted to encrypted
mmkv.reKey("Key_seq_1");

// change encryption key
mmkv.reKey("Key_seq_2");

// change from encrypted to unencrypted
kmmkv.reKey(null);
```

mrousavy:
Can you guys check if #314 works for you?

PR #314, opened 2022-02-07
feat: Re-Crypt an existing Storage instance (`storage.encrypt(..)`)

mrousavy:
```ts
const storage = new MMKV({ id: 'my.test.storage' })

storage.encrypt('my-private-key') // <-- encrypt with key 'my-private-key'
storage.set('test', 'works!')

storage.encrypt(undefined) // <-- remove encryption
storage.set('test', 'still works!')
```

closes: #179


ddikodroid:
so to encrypt, i need to call `storage.encrypt('key')` or `storage.recrypt('key')`?

in the README.md it stated like this
```
// encrypt all data with a private key
storage.recrypt('hunter2')

// remove encryption
storage.recrypt(undefined)
```

mrousavy:
recrypt, I renamed it - will release an update soon

PR #320, opened 2022-02-09
macos directory is missed in 2.1.0 after install

dimaportenko:
Directory `macos` is missed in the installed version in node_modules. I guess it is because `"macos/Mmkv.xcodeproj/project.pbxproj",` missed in the `files` section of the package.json. But I'm not much familiar how sources picked to the final package. @mrousavy please let me know if any other changes are required here.

CleanShot 2022-02-09 at 13 13 35


mrousavy:
Hey, thanks for the PR, nice catch!

I wonder if we need to roll two almost identical `.pbxproj` files though, can't we just have a single `apple/` folder that works both on iOS and MacOS?

mrousavy:
Hey Dima, have you found out anything yet?

mrousavy:
I think I've seen a library use an `apple/` folder instead before, it just required a small change in the react-native config file

dimaportenko:
hey @mrousavy thanks for heads up, not found a proper solution yet. On the current project, it works even without `macos` directory. My guess right now we don't need macos project file at all. It probably takes some time, but I'll recheck everything and back to you later.

dimaportenko:
Ok, I've retested everything and found out there is no need in `macos` directory at all. We just need to add `osx` target in the podspecs and that's it. 

I updated PR, now it just removes useless project file.

mrousavy:
Awesome, thanks!

PR #326, opened 2022-02-14
fix: Fix callback not passing new value

### LINKED ISSUE #324: Set new key value returns undefined value at listener
When I tried to use a listener,
```
useEffect(() => {
  const listener = storage.addOnValueChangedListener((changedKey) => {
    const newValue = storage.getString(changedKey)
    console.log(`"${changedKey}" new value: ${newValue}`)
  })
  // cleanup function
  return () => {
    listener.remove()
  }
}, [storage])
```
Setting new key value pairs would return undefined at newValue.
Updating existing key value pairs works fine.

Is this an expected behavior?

Thanks!

mrousavy:
* Fixes #324

### DIFF
--- a/src/MMKV.ts
+++ b/src/MMKV.ts
@@ -1,4 +1,3 @@
-import { unstable_batchedUpdates } from 'react-native';
 import { createMMKV } from './createMMKV';
 
 interface Listener {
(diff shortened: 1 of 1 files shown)


mrousavy:
Overruled by #386

mrousavy:
Overruled by #386

PR #278, opened 2022-01-04
[docs] zustand store and persist-store example added

Hirbod:
Added example how to use this lib with - imho - the best state-management lib "zustand"

PR #333, opened 2022-02-21
Convert zustand persist middleware example to Synchronous storage

gabrielmoncea:
Zustand persist middleware accepts synchronous storage methods too, so there is no need for a Promise based definition.
[Zustand Persist Docs](https://github.com/pmndrs/zustand/blob/main/src/middleware/persist.ts#L7)

### DIFF
--- a/docs/WRAPPER_ZUSTAND_PERSIST_MIDDLEWARE.md
+++ b/docs/WRAPPER_ZUSTAND_PERSIST_MIDDLEWARE.md
@@ -11,20 +11,14 @@ const storage = new MMKV()
 
 const zustandStorage: StateStorage = {
   setItem: (name, value) => {
-    storage.set(name, value)
-    return Promise.resolve()
+    return storage.set(name, value)
   },
   getItem: (name) => {
     const value = storage.getString(name)
-    if (value) {
-      return Promise.resolve(value)
-    } else {
-      return null
-    }
+    return value || null
   },
   removeItem: (name) => {
-    storage.delete(name)
-    return Promise.resolve()
+    return storage.delete(name)
   },
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/WRAPPER_ZUSTAND_PERSIST_MIDDLEWARE.md
```diff
@@ -11,20 +11,14 @@ const storage = new MMKV()
...
+    return storage.set(name, value)
   },
   getItem: (name) => {
     const value = storage.getString(name)
-    if (value) {
-      return Promise.resolve(value)
-    } else {
-      return null
-    }
+    return value || null
```
mrousavy, line 18:
```suggestion
    return value ?? null
```

mrousavy (changes requested):
Hey Gabriel, thanks for the PR! 

Just a quick nitpick, I prefer `??` over `||`, could you update it?

docs/WRAPPER_ZUSTAND_PERSIST_MIDDLEWARE.md
gabrielmoncea: @mrousavy if it returns an empty string then it will be an empty string instead of null.
```ts
'' ?? null // returns ''
'' || null // returns null
```

mrousavy: Yes empty string is desired here because it has a value - null is when it explicitly does not have a value

gabrielmoncea: @mrousavy it's done! I though that getString will always return a string, somehow without checking types :))

hirbod:
🙄 https://github.com/mrousavy/react-native-mmkv/pull/278/files

mrousavy:
@hirbod sorry I did not have enough time to review your PR at that time 😕
What are your thoughts on this PR?

hirbod:
I had kinda more details on how to use it, how to install it and how to have a persist/not persisted version of it

gabrielmoncea:
Hi @hirbod! I'm sorry for you PR. I took a look at it and might say that you definitely offer a bit more info. Although, the `getStorage` option doesn't work with current `zustand` api without `persist` middleware.

hirbod:
It's all good, I think what you provided is enough for most folks

PR #346, opened 2022-03-08
Fix: Main thread error with initialize function

umangloria:
It is mentioned in main MMKV package that  [MMKV initializeMMKV:nil] function should always initialize in a main thread.
Reference: https://github.com/Tencent/MMKV#quick-tutorial-1


mrousavy:
Hi, thank you for your PR!

1. Please indent the code correctly.
2. Please use `dispatch_sync` instead of async

umangloria:
Changes done. Please check now.

mrousavy:
great, thank you LGTM!

Issue #347, opened 2022-03-09
Path config with `~` on macOS seems to not work


Simek:
Lately, I have switched from AsyncStorage to MMKV in on of the macOS test projects, and I would like to utilize the `path` config, but it looks like that, when path includes tilde (`~`) file is not saved. Not sure if this is intended to work, but I have decided to report this anyway, so feel free to close this one or convert to feature request.

Other thing which was a bit unexpected is that you cannot use the constructor with config object, but without the `id` field (no fallback to default?), so for example when `path` and `encryption` field are the only present, this leads to error about missing ID.

# Repro
* Clone: https://github.com/Simek/react-native-macos-reminders
* Edit: https://github.com/Simek/react-native-macos-reminders/blob/master/src/Storage.js
* Add: `path: '~/.test'` config option

mrousavy:
Hey Bartosz!

1. I'll take a look! Maybe this issue should be reported in the original MMKV repo instead?
2. Yeah well ID cannot be default anymore since the default instance does not have encryption or custom path. So when you supply any of those, you have to use a different ID.

PR #350, opened 2022-03-10
chore: clarify the `id` field docs

Simek:
# Why

Refs:
* https://github.com/mrousavy/react-native-mmkv/issues/347#issuecomment-1063958167

# How

This PR adds a clarification to the `id` config field description and default fallback behavior, which has been explained by Marc in the referenced issue.


mrousavy:
Much better, thanks! 🖤

PR #384, opened 2022-04-19
fix getBoolean on the Web

eyousefifar:
local storage saves values in string format. so  Boolean('false') is always true. for this method to work correctly, it should check if value is 'true' or 'false' as string and returns the condition result.
to simplify the code I just return value === 'true'

### DIFF
--- a/src/createMMKV.web.ts
+++ b/src/createMMKV.web.ts
@@ -42,7 +42,15 @@ export const createMMKV = (config: MMKVConfiguration): NativeMMKV => {
     getBoolean: (key) => {
       const value = storage().getItem(key);
       if (value == null) return undefined;
-      return Boolean(value);
+      /**
+       * local storage saves keys and values in utf-16 dom strings
+       * @link https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage#description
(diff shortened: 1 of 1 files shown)


mrousavy:
Hey, thanks for the PR this is a good change, I missed that one!

src/createMMKV.web.ts
```diff
@@ -42,7 +42,15 @@ export const createMMKV = (config: MMKVConfiguration): NativeMMKV => {
...
       if (value == null) return undefined;
-      return Boolean(value);
+      /**
+       * local storage saves keys and values in utf-16 dom strings
+       * @link https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage#description
+       * true is saved as 'true', false as 'false'
+       * so if value == 'true' return true
+       * if value == 'false' return false
+       * I just simplified the conditions
+       */
```
mrousavy, line 52: Can you just remove this comment?

eyousefifar: yeah sure

eyousefifar: @mrousavy  done

mrousavy:
Great! Thanks for your PR 🙏

Issue #367, opened 2022-03-31
Jest Mocking react-native-mmkv


radelcom:
Hi,

is there a way or provided way to easily mock this great library? 

would be great if we can provide mocks directly from the library similar way to how [AsyncStorage](https://react-native-async-storage.github.io/async-storage/docs/advanced/jest) does

Thanks!

mrousavy:
Hey, this seems to be very weird to let the user set up all of this - I think there's better examples (e.g. Reanimated) that automatically mock everything

radelcom:
@mrousavy, yeah that looks simpler. Anything that will help other people using this package so that we dont need to look else where or make our own mocking.

mrousavy:
Sure, if you shoot me a PR I'll be happy to review, test and merge it :)

mrousavy:
Hey! I think mocking should now work! Thanks everyone! :)

norellanac:
Add a wrapper provider to your App/Component

```
import { Provider } from 'react-redux';
import store from '../stores/store';

export const renderWithProviderTheme = (component: React.ReactElement) => {
  const { debug, ...rest } = render(
      <Provider store={store}>{component}</Provider>,
  );

  return {
    debug,
    ...rest,
  };
};
```

**********************
```
import { renderWithProviderTheme } from '...';

describe('Render with provider, store(mmk)', () => {
  test('Should render correctly component', () => {
    renderWithProviderTheme(
      <MyComponent/>,
    );
  });
```

PR #395, opened 2022-05-05
Automocking in test environment

ruibinch:
When running in a test environment, use a mock MMKV instance that utilises an underlying `Map` structure. This allows tests on this package to work without having to manually define mocks or any jest setup files.

Inspired by #367.

### DIFF
--- a/src/MMKV.ts
+++ b/src/MMKV.ts
@@ -1,4 +1,6 @@
 import { createMMKV } from './createMMKV';
+import { createMockMMKV } from './createMMKV.mock';
+import { isJest } from './PlatformChecker';
 
 interface Listener {
   remove: () => void;
@@ -134,7 +136,9 @@ export class MMKV implements MMKVInterface {
    */
   constructor(configuration: MMKVConfiguration = { id: 'mmkv.default' }) {
     this.id = configuration.id;
-    this.nativeInstance = createMMKV(configuration);
+    this.nativeInstance = isJest()
+      ? createMockMMKV()
+      : createMMKV(configuration);
     this.functionCache = {};
   }
 
--- a/src/PlatformChecker.ts
+++ b/src/PlatformChecker.ts
@@ -1,0 +1,3 @@
+export function isJest(): boolean {
+  return process.env.JEST_WORKER_ID != null;
+}
--- a/src/createMMKV.mock.ts
+++ b/src/createMMKV.mock.ts
@@ -1,0 +1,29 @@
+import type { NativeMMKV } from 'react-native-mmkv';
+
+/* Mock MMKV instance for use in tests */
+export const createMockMMKV = (): NativeMMKV => {
+  const storage = new Map<string, string | boolean | number>();
+
+  return {
+    clearAll: () => storage.clear(),
+    delete: (key) => storage.delete(key),
+    set: (key, value) => storage.set(key, value),
+    getString: (key) => {
+      const result = storage.get(key);
+      return typeof result === 'string' ? result : undefined;
+    },
+    getNumber: (key) => {
+      const result = storage.get(key);
+      return typeof result === 'number' ? result : undefined;
+    },
+    getBoolean: (key) => {
+      const result = storage.get(key);
+      return typeof result === 'boolean' ? result : undefined;
+    },
+    getAllKeys: () => Array.from(storage.keys()),
+    contains: (key) => storage.has(key),
+    recrypt: () => {
+      console.warn('Encryption is not supported in mocked MMKV instances!');
+    },
+  };
+};
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,4 +1,5 @@
 {
+  "exclude": ["example"],
   "compilerOptions": {
     "baseUrl": ".",
     "paths": {
--- a/README.md
+++ b/README.md
@@ -170,6 +170,10 @@ storage.recrypt('hunter2')
 storage.recrypt(undefined)
 ```
 
+## Testing with Jest
+
+A mocked MMKV instance is automatically used when testing with Jest, so you will be able to use `new MMKV()` as per normal in your tests. Refer to [example/test/MMKV.test.ts](example/test/MMKV.test.ts) for an example.
+
 ## Documentation
 
 * [Hooks](./docs/HOOKS.md)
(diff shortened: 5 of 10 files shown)


src/MMKV.ts
```diff
@@ -132,9 +134,15 @@ export class MMKV implements MMKVInterface {
    * Creates a new MMKV instance with the given Configuration.
    * If no custom `id` is supplied, `'mmkv.default'` will be used.
    */
-  constructor(configuration: MMKVConfiguration = { id: 'mmkv.default' }) {
+  constructor(
+    configuration: MMKVConfiguration = {
+      id: isJest() ? 'mmkv.mock' : 'mmkv.default',
```
mrousavy, line 139: This doesn't seem to be required. Can we remove that change?

src/PlatformChecker.ts
```diff
@@ -0,0 +1,3 @@
+export function isJest(): boolean {
+  return Boolean(process.env.JEST_WORKER_ID);
```
mrousavy, line 2: Wouldn't a `process.env.JEST_WORKER_ID != null` be more semantically correct? Does the linter say anything about this?

src/createMMKV.mock.ts
```diff
@@ -0,0 +1,29 @@
+/* Mock MMKV instance for use in tests */
```
mrousavy, line 1: Move that comment down to line 3 so it documents that function

mrousavy (changes requested):
Hey, thanks for your PR, this is awesome!

1. I left a few comments, once we got those resolved we're good otherwise. Code looks clean!
2. Can you confirm that with this change you can now use MMKV without any issues in both actually running the app (no broken behaviour) as well as when running Jest and MMKV instances are being accessible?

Beyond this PR:

1. Do you think we could add a GitHub action (see .github/workflows folder) that runs the Jest tests? Not sure if that really is needed since it doesn't test the actual native MMKV stuff but only the Jest wrapper, but still
2. Do you think it would be possible to test native MMKV functionality as if it was running in an actual RN app with a GitHub action? Maybe spin up a virtual device or something to see if everything's working?

mrousavy:
also CI is failing:

Screenshot 2022-05-05 at 14 10 05

src/PlatformChecker.ts
ruibinch: No linter complaints, but agreed that `!= null` captures the meaning better.

ruibinch:
Resolved the comments; let me test this under more situations tomorrow to confirm that everything remains good and update again.

Re other questions:

1. Sounds good, let me add that in a future PR.
2. I've had minimal exposure to GH actions so no idea currently on its capabilities and whether it can support an emulator, but sounds interesting and worth a deeper investigation!

ruibinch:
Tested a couple of scenarios:

1. MMKV continues to work as per normal in an app
2. Running tests, with an open app running an MMKV instance, does not break anything (data remains in the app's MMKV instance)

src/MMKV.ts
```diff
@@ -1,4 +1,6 @@
 import { createMMKV } from './createMMKV';
+import { createMockMMKV } from './createMMKV.mock';
```
shamilovtim, line 2: I assume you need to use the `require()` API to import this behind an if statement, otherwise you're including it in production builds

src/MMKV.ts
```diff
@@ -1,4 +1,6 @@
...
+import { isJest } from './PlatformChecker';
```
shamilovtim, line 3: having an export for 1 LOC seems unnecessary. also, file is capitalized despite not being a react component or class.

src/MMKV.ts
ruibinch: We need to include this in production builds in order for the automocking to work right?

src/MMKV.ts
ruibinch: Understandable, but I feel this is cleaner than having `process.env.JEST_WORKER_ID != null` inside the `MMKV` constructor.

Re filename - I viewed `PlatformChecker` as more of a "sub-domain" and not a generic file like `helpers`/`utils` hence the capitalisation; not a strong preference though, happy to change it to something else if that's the concensus.

src/MMKV.ts
shamilovtim: What I'm saying is you're going to include this file in people's JS bundles unless you use the require API and but it behind an if statement

src/MMKV.ts
mrousavy: I like it the way it is, let's leave it.

src/MMKV.ts
mrousavy: Is that a fact? We can lazy-load it at the bottom if needed?

mrousavy: I mean technically it's not really impacting bundle size, it's just a few lines and no extra dependencies

tsconfig.json
```diff
@@ -1,4 +1,5 @@
 {
+  "exclude": ["example"],
```
mrousavy, line 2: Why did you add this here? Now the `example/` doesn't do any typechecking because it doesn't have it's own tsconfig, no?

ruibinch: I added a `example/tsconfig.json` file; `yarn tsc` at the root level throws type errors when it type-checks the `example/**/*.test.ts` files because the Jest types are only defined in `example/node_modules`, so figured it's neater to have separate tsconfig files.

ruibinch: Just realised that the "Run TypeScript in example/" job wasn't running because `typescript` script wasn't defined in `example/package.json` - updated that

mrousavy:
woohoo!

mrousavy:
Released [2.4.0](https://github.com/mrousavy/react-native-mmkv/releases/tag/v2.4.0) 🚀  - thanks for your PR!

ruibinch:
Installing 2.4.0 doesn't include the code changes in `/lib`, though running `yarn build` locally on the current master branch does contain the code changes. Is there some additional config or something else required?

ruibinch:
Cheers, all good now!

AshciR:
This was a joy to read. 

I'm new to mobile development, and I'm learning the ecosystem. I appreciate you enabling developers to write tests for your library.

jahirfiquitiva:
@ruibinch @mrousavy this added an example for a single test... would you mind providing an example or guiding me on how to write multiple tests in multiple files that refer a single mmkv instance shared across all the tests?

PR #402, opened 2022-06-02
IOS/app-group

RZulfikri:
Hi @mrousavy, thanks for making an awesome repo, it really helps and is easy to understand. Just wanna contribute a bit to adding app-group options for IOS.  App group used to communicate between `MMKVCore` and `MMKVExtension` or `MMKVWatchExtension`. In my case, I need an app-group to implement the IOS Share extension.


mrousavy:
Hi, thanks for your PR!

`appGroup` is a new field that needs to be added to the types and documented so it can be used. Other than that it looks good to me!

RZulfikri:
@mrousavy I've updated the types and documentation, if you have any time please review and give me feedback.  thanks.

README.md
```diff
@@ -115,6 +116,7 @@ The following values can be configured:
...
 * `path`: The MMKV instance's root path. By default, MMKV stores file inside `$(Documents)/mmkv/`. You can customize MMKV's root directory on MMKV initialization (documentation: [iOS](https://github.com/Tencent/MMKV/wiki/iOS_advance#customize-location) / [Android](https://github.com/Tencent/MMKV/wiki/android_advance#customize-location))
 * `encryptionKey`: The MMKV instance's encryption/decryption key. By default, MMKV stores all key-values in plain text on file, relying on iOS's/Android's sandbox to make sure the file is encrypted. Should you worry about information leaking, you can choose to encrypt MMKV. (documentation: [iOS](https://github.com/Tencent/MMKV/wiki/iOS_advance#encryption) / [Android](https://github.com/Tencent/MMKV/wiki/android_advance#encryption))
+* `appGroup`: The MMKV instance's app group name `(IOS Only)`. It is used to share MMKV instances between apps and extensions within the app group. (documentation: [iOS](https://github.com/Tencent/MMKV/wiki/iOS_tutorial#configuration))
```
mrousavy, line 119: 1. "IOS Only" -> "iOS only"
2. Remove backticks (`) for the iOS only thing. It's not a code block.
3. "within the app group" -> "within the same app group"
4. "documentation: IOS" -> "documentation: iOS"

mrousavy (changes requested):
Added some very minor nitpicks, otherwise good to go!

ios/MmkvHostObject.mm
```diff
@@ -10,10 +10,16 @@
...
-MmkvHostObject::MmkvHostObject(NSString* instanceId, NSString* path, NSString* cryptKey)
+MmkvHostObject::MmkvHostObject(NSString* instanceId, NSString* path, NSString* cryptKey, NSString* appGroup)
 {
+  NSString *groupDir = [[NSFileManager defaultManager] containerURLForSecurityApplicationGroupIdentifier:appGroup].path;
   NSData* cryptData = cryptKey == nil ? nil : [cryptKey dataUsingEncoding:NSUTF8StringEncoding];
-  instance = [MMKV mmkvWithID:instanceId cryptKey:cryptData rootPath:path];
+  if (appGroup == nil) {
+    instance = [MMKV mmkvWithID:instanceId cryptKey:cryptData rootPath:path];
+  } else {
+    [MMKV initializeMMKV:nil groupDir:groupDir logLevel:MMKVLogNone];
```
mrousavy, line 20: Hmm, this causes MMKV to re-initialize. Is that a safe thing to do? Right now we call `initializeMMKV` at the launch of the app... Maybe that's when the appGroup needs to be specified, instead of everytime you do new MMKV()? 🤔

RZulfikri: ah, I saw what you mean. I'll try to move and test it again. For now, it's working, but I don't know the side effect.

RZulfikri: @mrousavy btw if thats the cases, it seems we cannot put appGroup as a parameters in JS side. I'll try to detect in Native side instead of JS.

RZulfikri:
@mrousavy I change AppGroup parameters from the JS side to the Native side by putting the **AppGroup** key in info.plist

Screenshot 2022-06-15 at 07 14 49

mrousavy:
That looks like a good solution. We need to create a doc that explains this instead of just linking to this PR though. I'll get to that after I'm back from vacation :)

Flam3rboy:
Any updates on this?

shawarmaz:
Is it just docs that are pending? i.e. is the code itself good to go?

mrousavy:
hey sorry - I'll try to take another look at that soon

RZulfikri:
@mrousavy, It been so long. Mention me if you need anything related to this PR, I'm ready to help.

theLaurensS:
@mrousavy What can we do to help to get this PR approved?

Is it as simple as updating the docs like this? https://github.com/RZulfikri/react-native-mmkv/pull/1/files 
(I created PR to @RZulfikri his branch)

mrousavy:
Do you think it makes sense to make appGroup a field you can pass in from JS? I think that should be possible if we just call initializeMMKV again, let me find out if that has any side-effects

RZulfikri:
@mrousavy I dont think it is good to put on JS, because later when you’re try to implement mmkv in the extension you will need the app group on native code. So, instead of put on both (js and native/extension) I think better to put on native only.

mrousavy:
One question; Why don't we just always use AppGroup? Is there a downside here?

SamuelScheit:
I'm not sure, but I think AppGroups can't sync to iCloud.

README.md
```diff
@@ -115,6 +115,7 @@ The following values can be configured:
...
 * `path`: The MMKV instance's root path. By default, MMKV stores file inside `$(Documents)/mmkv/`. You can customize MMKV's root directory on MMKV initialization (documentation: [iOS](https://github.com/Tencent/MMKV/wiki/iOS_advance#customize-location) / [Android](https://github.com/Tencent/MMKV/wiki/android_advance#customize-location))
 * `encryptionKey`: The MMKV instance's encryption/decryption key. By default, MMKV stores all key-values in plain text on file, relying on iOS's/Android's sandbox to make sure the file is encrypted. Should you worry about information leaking, you can choose to encrypt MMKV. (documentation: [iOS](https://github.com/Tencent/MMKV/wiki/iOS_advance#encryption) / [Android](https://github.com/Tencent/MMKV/wiki/android_advance#encryption))
+* `appGroup`: The MMKV instance's app group name (iOS only). It is used to share MMKV instances between apps and extensions within the same app group. (documentation: [iOS](https://github.com/Tencent/MMKV/wiki/iOS_tutorial#configuration))
```
mrousavy, line 118: This shouldn't belong here, as it's not a prop in `MMKV.ts`'s constructor.

There should be a separate section in the README explaining how to run it under an AppGroup (with a disclaimer that `path` won't work then)

ios/MmkvHostObject.mm
```diff
@@ -28,11 +35,13 @@
...
-    bool pathExists = [fileManager fileExistsAtPath:path isDirectory:nil];
-    if (!pathExists) {
-      throw std::runtime_error("Failed to create MMKV instance! The given Storage Path does not exist!");
+    if (appGroup == nil) {
+      // Check if path is invalid
+      NSFileManager* fileManager = [[NSFileManager alloc] init];
+      bool pathExists = [fileManager fileExistsAtPath:path isDirectory:nil];
+      if (!pathExists) {
+        throw std::runtime_error("Failed to create MMKV instance! The given Storage Path does not exist!");
+      }
```
mrousavy, line 44: Maybe we could also warn to console if the user supplied a custom path and we're using appgroups? Because then path wont work

mrousavy (changes requested):
Okay did another pass at the review. One more thing; maybe we could also add that to the Expo Config Plugin?

Jokerwolf:
Hey @RZulfikri, good PR, really looking forward to this change too, by any chance you could resolve the comments anytime soon please? 😊

RZulfikri:
merged here https://github.com/mrousavy/react-native-mmkv/pull/546

Issue #299, opened 2022-01-22
android cmake


wq13718048209:
"react-naitve":"0.63.4"
"react-native-mmkv": "^2.0.0",
My native cmake version is 3.9.2, but it is Android./gradlew build will cause an error  ( CMake 3.9.0 or higher is required.  You are running version 3.6.0-rc2) ,
I don't have the 3.6.0-Rc2 version of Cmake locally, but it will always be installed automatically after bulid ends and an error will be reported,
I tried deleting it, but as soon as the build finished it would come up and give an error. I felt like I was stuck in a loop

mrousavy:
Maybe NDK Version is missing in your app's build.gradle

PR #406, opened 2022-06-07
Force cmake >3.9.0 to fix build issues

JBakerSitesafe:
This project requires at least cmake version 3.9.0 to compile properly, while Android Studio's sdk manager defaults to cmake 3.6.0.

Despite updating cmake both through Android Studio and manually, and configuring my project to use the new versions, the build process was still using 3.6.0. Setting the cmake version and ndk version in my app's build,gradle android along with cmake.dir in gradle.properties did not help.

I believe this is the same issue as #299 

This simple change fixed it on my end, without requiring the library user to hack around with configurations until it works.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -113,6 +113,7 @@ android {
 
   externalNativeBuild {
     cmake {
+      version "3.9.0+"
       path "CMakeLists.txt"
     }
   }


JBakerSitesafe:
Another fix was upgrading gradle build tools from 3.5.3 to 4.2.2

mrousavy:
Cool! Let's get to master

mrousavy:
Actually - huh - the CI fails to build with your change...

Screenshot 2022-06-13 at 10 24 43

JBakerSitesafe:
Strange; builds fine on my machine, must be a difference in configurations somewhere. I suppose if anyone else has the same issue they should try the change manually themselves or update gradle.

PR #408, opened 2022-06-14
docs: add Jotai wrapper example

sregg:
Add docs about using MMKV to persist Jotai atoms.


mrousavy:
Hi Simon, nice seeing you in the wild!

Do Jotai users know how to use that function? If not, maybe some additional "usage guide" would help to teach them how to use that? Something like "Then simply use `atomWithStorage` to create your atoms: [code]"

sregg:
That's a good idea.
I can add a link to the official `atomWithStorage` doc: https://jotai.org/docs/utils/atom-with-storage

lpatrun:
Hello, do you happen to know how to test atomWithStorage? I keep getting the same error. I tried a couple of ideas but nothing worked. My error
`● Test suite failed to run

    TypeError: (0 , atomWithStorage.atomWithStorage) is not a function

      1 | import { atomWithStorage } from './atomWithStorage';
      2 |
    > 3 | export const introShown = atomWithStorage('introShown', false);`

PR #419, opened 2022-06-30
feat: make hook reactive when instance change

DCKT:
Hello,

The code is not production ready, but the idea is here, for better performance we should use the **instance id** in order to compare correctly the instance.

The id field is private actually so i don't know if we would like to expose it ?


mrousavy:
Hi! I think I implemented this in #422, is your PR still valid?

DCKT:
Hi, it's good with yours :)

PR #451, opened 2022-09-14
fix: use correct function call in jotai wrapper example

fobos531:
Hey @mrousavy , thanks for the awesome library!

I just have a small change, which corrects the jotai wrapper example to use the correct function call in respect to the imports from `jotai/utils` 

Thanks!


mrousavy:
Thanks! LGTM

PR #454, opened 2022-09-16
Example fixes

riledv:
"Type 'string | undefined' is not assignable to type 'string | T | Promise<string | T | null> | null'."


mrousavy:
Thanks! Can you replace || with ?? please?

haresh12 (approved):

PR #461, opened 2022-10-11
feat: Implement `trim()` to remove unused space

mrousavy:


ken0nek:
Hi @mrousavy,
always appreciate your work!

Do you have any estimate for releasing it?

android/src/main/cpp/MmkvHostObject.cpp
```diff
@@ -221,6 +221,20 @@ jsi::Value MmkvHostObject::get(jsi::Runtime& runtime, const jsi::PropNameID& pro
     });
   }
 
+  if (propName == "trim") {
+    // MMKV.trim()
```
ken0nek, line 225: Just reference
* [MMKV/MMKV.h at master · Tencent/MMKV](https://github.com/Tencent/MMKV/blob/master/Core/MMKV.h#L310-L313)
* [MMKV/MMKV\_IO.cpp at master · Tencent/MMKV](https://github.com/Tencent/MMKV/blob/master/Core/MMKV_IO.cpp#L1008-L1046)

mrousavy:
I think `trim()` doesn't work the way I thought it would work. Try it for yourself

android/src/main/cpp/MmkvHostObject.cpp
mrousavy: wdym for reference?

mrousavy:
Closing as superseded: trim() is already implemented on current main.

PR #524, opened 2023-03-04
feat: Support `id` on web

### LINKED ISSUE #520: Feature request: 'id' support on Web
When trying to pass the `id` option to `createMMKV` on the web, it [throws an exception](https://github.com/mrousavy/react-native-mmkv/blob/f3dc49b55c0fa0c1440f4e0aaf308125913b6ff3/src/createMMKV.web.ts#LL10C28-L10C56) about this not being supported.

I realise that there can't really be truly separate stores on web because of how `localStorage` works, but it could be reasonably emulated (and the caveats about that emulation noted in the documentation) by namespacing the keys.

I'd be happy to contribute this feature, but I'd like to check if there was a reason it wasn't implemented initially that I might not have thought about.

jonathanj:
**I'm putting this PR up as a draft because I still need to put these changes through its paces in our codebase, but I would appreciate any feedback you might have in the interim.**

# Description
Local storage does not have a concept of separate stores in user code, this change emulates that behaviour by prefixing the keys in local storage with `id`.

This retains backwards compatibility with existing web MMKV stores by treating the the default `mmkv.default` id is as prefixless.

Fixes #520

# Questions I have
- **How can I exercise the web code in the unit tests?**
   The unit tests don't seem to exercise the `id` option, I'm open to extending the test coverage for this, but for this change it would probably also make sense to actually exercise the web code in tests.
- **Running `yarn lint` fails unexpectedly**
  ```
  Error: package.json » @react-native-community/eslint-config#overrides[2]:
        Environment key "jest/globals" is unknown
  ```
  It does fail on `master` like this too for me, so I'm not sure if this an issue with my local environment.

### DIFF
--- a/src/createMMKV.web.ts
+++ b/src/createMMKV.web.ts
@@ -29,6 +29,10 @@ export const createMMKV = (config: MMKVConfiguration): NativeMMKV => {
 
   const textEncoder = createTextEncoder();
 
+  if (config.id.indexOf('\\') !== -1) {
+    throw new Error('MMKV: `id` cannot contain the backslash character (`\\`)!');
+  }
+
   const keyPrefix = (config.id && config.id !== 'mmkv.default') ? `${config.id}\\` : null;
   const prefixedKey = (key: string) => {
     if (keyPrefix === null) return key;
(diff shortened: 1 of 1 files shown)


mrousavy:
Hi!

We need a wildcard system for this to work, so e.g. use one special character/character-combination (`\\`) to separate namespace (id) and key value.

So: `MMKV(id: 'test')`:

* `'some-key'` -> `'test\\some-key'
* `'some-\\-key'` -> `throw new Error("Cannot use wildcard `\\` in key!")`

And `MMKV()`:

* `'some-key'` -> `'some-key'
* `'some-\\-key'` -> `throw new Error("Cannot use wildcard `\\` in key!")`

jonathanj:
@mrousavy Thanks for the feedback, and sorry about the long delay in replying. I've updated the PR to throw when trying to use a backslash in the `id` or `key`.

mrousavy:
Cool! Thanks for your PR! I'll do some minor cleanup in a follow up commit, but looks good overall! LGTM

mrousavy:
Released in 2.8.0! 🎉

jonathanj:
Thank you, @mrousavy!

PR #532, opened 2023-04-07
fix: Clean up `id` on Web

mrousavy:
a bit cleaner


mrousavy:
cc @jonathanj (we forgot about clearAll as well)

jonathanj:
Good catch, thanks for adding that! 🙏

PR #533, opened 2023-04-07
feat: add in-memory storage as a fallback for disabled LocalStorage

jpudysz:
Hello @mrousavy ✋ 

I am submitting this pull request to propose the addition of a new feature `webFallbackStorage`. The purpose of this enhancement is to provide a more resilient storage option for web environments, particularly when users disable `localStorage` in their browsers.

The `webFallbackStorage` is designed as an optional property that conforms to the `Storage` type. It allows the application to handle disabled `localStorage` situations gracefully by offering eg. in-memory storage as an alternative. This ensures smooth functioning of the application despite any user-imposed limitations.

Right now, `react-native-mmkv` throws an error and there is no way to run the `monorepo` with disabled `localStorage`.

Here's a brief overview of the changes introduced with this feature:

* The `webFallbackStorage` property has been added as an optional attribute, maintaining backward compatibility with existing implementations.
* When a user disables `localStorage`, the application will seamlessly transition to using `webFallbackStorage`, ensuring no disruption in the user experience.
* If the `webFallbackStorage` property is not provided, the default behavior remains unchanged, and an error will be thrown in case `localStorage` is disabled.

I believe this enhancement will significantly improve the robustness and flexibility of the `react-native-mmkv` library.
Later we can add default `in-memory` storage to the lib.

Please feel free to reach out to me with any questions or suggestions.


jpudysz:
This is default behaviour:
White screen + such error message in the console:
err

README.md
```diff
@@ -182,6 +184,101 @@ const buffer = storage.getBuffer('someToken')
...
+```
+
+### Buffers
+
+```js
+storage.set('someToken', new Uint8Array([1, 100, 255]))
+const buffer = storage.getBuffer('someToken')
+console.log(buffer) // [1, 100, 255]
+```
+
```
mrousavy, line 254:
```suggestion
```

README.md
```diff
@@ -115,6 +116,7 @@ The following values can be configured:
...
 * `encryptionKey`: The MMKV instance's encryption/decryption key. By default, MMKV stores all key-values in plain text on file, relying on iOS's/Android's sandbox to make sure the file is encrypted. Should you worry about information leaking, you can choose to encrypt MMKV. (documentation: [iOS](https://github.com/Tencent/MMKV/wiki/iOS_advance#encryption) / [Android](https://github.com/Tencent/MMKV/wiki/android_advance#encryption))
+* `webFallbackStorage`: An alternative storage solution conforming to the `Storage` type. This is utilized as a fallback for web environments when, for instance, a user disables `localStorage` in their browser. This feature enables your application to manage such situations gracefully, offering eg. in-memory storage as an alternative (see example below). Please note that this property is optional; by default, an error will be thrown in case of a disabled localStorage.
```
mrousavy, line 119: Do you really think we should provide a custom fallback here? I feel like the user should provide a fallback for the entire MMKV thing if he really wants to use a different lib. Maybe we can just add a prop like "`fallbackToInMemoryMap: boolean`" which just uses a local Map and doesnt write to storage? (With a proper disclaimer)

src/createMMKV.web.ts
```diff
@@ -21,6 +21,16 @@ export const createMMKV = (config: MMKVConfiguration): NativeMMKV => {
         'Tried to access storage on the server. Did you forget to call this in useEffect?'
       );
     }
+
+    try {
+      // throws ACCESS_DENIED error
+      window.localStorage;
+    } catch {
+      if (config.webFallbackStorage) {
+        return config.webFallbackStorage;
```
mrousavy, line 30: re my prev comment; here would be the point where we construct the dummy `Map<K,V>` storage

mrousavy (changes requested):
hey! thanks for your PR man! Left some comments, lmk :)

README.md
jpudysz: Good point. I was also wondering if we need to ask the user to provide a fallback, or if it's too much.

But, what if we don't ask the user anything? It will be enabled by default. We can console.warn a message like DataDog does:

image

Something like:
```
Local storage is disabled your experience will be limited to in-memory storage
```

If you think about it, who would want to get a crash (especially on production)? The library can provide be default a `Map`, and maybe for the convenience of the developer, return some flag to be consumed later to display some kind of message/popup like Google does, for example:

image

It's still better than a crash.

I was wondering about something like:

```typescript
return {
    clearAll: () => {},
    delete: (key) => {},
    set: (key, value) => {},
    //  (...) other functions,
    inMemoryFallback: boolean // for web depends on localStorage permission, for native always true
```

Let me know what do you think about it

mrousavy: Yea I think this is better. I'm not sure if I want that flag though, this is a bit confusing for everyone that didn't think of this behaviour... Maybe we want web devs to intentionally check for localStorage?

jpudysz:
Ready to be checked @mrousavy

warn

We have a single warning and my app is up and running 🪄

mrousavy:
Hey - for some reason the linter failed. I'll fix that, your PR looks good!

I'm thinking if we should also use the in-memory Map fallback for when JSI is unavailable, e.g. when running in a Chrome debugger?

jpudysz:
> Hey - for some reason the linter failed. I'll fix that, your PR looks good!
> 
> I'm thinking if we should also use the in-memory Map fallback for when JSI is unavailable, e.g. when running in a Chrome debugger?

If it causes a crash then yes, it's still better to have limited experience, yet let the user debug the app.

hirbod:
I think the console.log should only be printed if `typeof window !== "undefined"`

PR #534, opened 2023-04-07
Create WRAPPER_RECOIL

dezudas:
For recoil persist we can use the following code.

### DIFF
--- a/docs/WRAPPER_RECOIL
+++ b/docs/WRAPPER_RECOIL
@@ -1,0 +1,24 @@
+# recoil storage wrapper
+
+If you want to use MMKV with [recoil](https://recoiljs.org/), Use the following persist code:
+
+```tsx
+const persistAtom = (key) => ({ setSelf, onSet }) => {
+  setSelf(() => {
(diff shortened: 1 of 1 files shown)


docs/WRAPPER_RECOIL
```diff
@@ -0,0 +1,24 @@
...
+
+```tsx
+const persistAtom = (key) => ({ setSelf, onSet }) => {
+  setSelf(() => {
+    let data = storage.getString(key);
+    if(data != null){
+      return JSON.parse(data);
+    }else {
+      return new DefaultValue();
+    }
```
mrousavy, line 13:
```suggestion
    let data = storage.getString(key);
    if (data != null){
      return JSON.parse(data);
    } else {
      return new DefaultValue();
    }
```

mrousavy (changes requested):
Hey, thanks for your PR! Looks good, just a quick formatting fix

docs/WRAPPER_RECOIL
dezudas: looks good

PR #535, opened 2023-04-11
feat: Batching

mrousavy:
```ts
const storage = new MMKV()

storage.batch((s) => {
  s.set('hello', 'world')
  s.set('test', 'another')
  s.set('someLargeBuffer', buffer)
})
```

### DIFF
--- a/ios/MmkvHostObject.mm
+++ b/ios/MmkvHostObject.mm
@@ -295,6 +295,38 @@ jsi::Value MmkvHostObject::get(jsi::Runtime& runtime, const jsi::PropNameID& pro
       return jsi::Value::undefined();
     });
   }
+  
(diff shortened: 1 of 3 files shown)


adhorodyski:
Love the feature! 
I've been playing with the library lately for https://github.com/Expensify/App and have a quick question - as far as I understand this will trigger `addOnValueChangedListener` for each updated key (20k keys when eg. populating will do 20k events etc), do you think it might be worth to have a separate listener notifying about the batched updates being applied? Like a `addOnBatchedValuesChangedListener` 😅 that will trigger only once.

mrousavy:
Hey - yea for listeners we should maybe batch that as well then. 🤔

Issue #500, opened 2023-01-17
Android build issue on 2.6.1 with react-native 0.70.6


ridvanaltun:
Versions:
- react-native version **0.70.6**
- react-native-mmkv version **2.6.1**

<details>
	<summary>Click to expand error</summary>

```
ViewCommandHelper$ScrollToEndCommandData found in modules jetified-react-android-0.71-release.0-release-runtime (com.facebook.react:react-android:0.71.0) and jetified-react-native-0.70-release.6-release-runtime (com.facebook.react:react-native:0.70.6)
     Duplicate class com.facebook.react.views.scroll.ReactScrollViewHelper found in modules jetified-react-android-0.71-release.0-release-runtime (com.facebook.react:react-android:0.71.0) and jetified-react-native-0.70-release.6-release-runtime (com.facebook.react:react-native:0.70.6)
     Duplicate class com.facebook.react.views.scroll.ReactScrollViewHelper$1 found in modules jetified-react-android-0.71-release.0-release-runtime (com.facebook.react:react-android:0.71.0) and jetified-react-native-0.70-release.6-release-runtime (com.facebook.react:react-native:0.70.6)
     Duplicate class com.facebook.react.views.scroll.ReactScrollViewHelper$2 found in modules jetified-react-android-0.71-release.0-release-runtime (com.facebook.react:react-android:0.71.0) and jetified-react-native-0.70-release.6-release-runtime (com.facebook.react:react-native:0.70.6)
     Duplicate class com.facebook.react.views.scroll.ReactScrollViewHelper$HasFlingAnimator found in modules jetified-react-android-0.71-release.0-release-runtime (com.facebook.react:react-android:0.71.0) and jetified-react-native-0.70-release.6-release-runtime (com.facebook.react:react-native:0.70.6)
     Duplicate class com.facebook.react.views.scroll.ReactScrollViewHelper$HasScrollEventThrottle found in modules jetified-react-android-0.71-release.0-release-runtime (com.facebook.react:react-android:0.71.0) and jetified-react-native-0.70-release.6-release-runtime (com.facebook.react:react-native:0.70.6)
     Duplicate class com.facebook.react.views.scroll.ReactScrollViewHelper$HasScrollState found in modules jetified-react-android-0.71-release.0-release-runtime (com.facebook.react:react-android:0.71.0) and jetified-react-native-0.70-release.6-release-runtime (com.facebook.react:react-native:0.70.6)
[...]

PR #544, opened 2023-06-01
docs: note about `react-native` <-> `react-native-mmkv` versions compativility

vshkl:
I needed to use this library in a project with a rather old version of `react-native` (`v0.65.3`, for that matter). Of course, I defaulted to the latest version because the documentation didn't tell me otherwise. Only after some time I found out that in my case I can't use the latest version. Other people apparently sometimes find this information the hard way, as in this [issue](https://github.com/mrousavy/react-native-mmkv/issues/500). So I decided it wouldn't hurt to make a note of it in the project README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,8 @@ yarn add react-native-mmkv
 cd ios && pod install
 ```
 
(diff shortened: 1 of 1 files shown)


mrousavy:
Closing as stale: this compatibility note is outdated for the current major version and docs.

PR #546, opened 2023-06-07
iOS/app-group MMKV MultiProcess mode

Jokerwolf:
Replicating existing PR https://github.com/mrousavy/react-native-mmkv/pull/402 and following on open comments as would like to use this change in my code.


mrousavy:
Awesome! Thanks for your PR, this looks pretty good - I think I'm gonna follow up with an expo config plugin and then we can release it.

Can you confirm that it works both with AppGroup enabled and with a Path passed?

Jokerwolf:
Perfect! Thank you 🙇

Just revalidated: if both `path` and `AppGroup` are set there will be a warning in the output. The application meanwhile will start and run normally reading/writing to MMKV storage.
image

mrousavy:
Hey sorry man my bad, let's merge this thanks so much for your contribution!

jamesedmonston:
This looks great! @mrousavy are you still planning to implement a Config Plugin?

PR #550, opened 2023-06-19
feat(android): add support for upcoming react native 0.73

ddikodroid:
As per https://github.com/react-native-community/discussions-and-proposals/issues/671, this change is needed to support a future release of React Native.

PR #553, opened 2023-06-21
feat(android): add support RN <0.71 when use namespace

MasonLe2497:
In the [PR](https://github.com/mrousavy/react-native-mmkv/pull/550), it drop support RN <0.71.
In this PR, i remove package from AndroidManifest.xml file when ANDROID_GRADLE_PLUGIN_VERSION >=7

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-native-mmkv",
-  "version": "2.10.0",
+  "version": "2.10.1",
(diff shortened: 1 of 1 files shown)


mrousavy:
Is there any harm in just adding `namespace` no matter if we're on RN 0.70 or RN 0.73? Why do we need to check that?

mrousavy:
Great - @MasonLe2497 can you update your PR to revert all changes in build.gradle and just have the updated AndroidManifest here? Thanks!

PR #556, opened 2023-06-23
fix: prevent console.warn spam in server environment

hirbod:
The latest version 2.10.0 added a good fallback strategy for browsers with disabled localStorage. The warning is also useful, but it tends to spam and flood the console/logs of Next.js/Node server environments. To address this, you can add a simple check to limit the console.log statement to the browser environment only.

Addresses a regression caused by https://github.com/mrousavy/react-native-mmkv/pull/533

### DIFF
--- a/src/createMMKV.web.ts
+++ b/src/createMMKV.web.ts
@@ -26,8 +26,9 @@ export const createMMKV = (config: MMKVConfiguration): NativeMMKV => {
   if (config.path != null) {
     throw new Error("MMKV: 'path' is not supported on Web!");
   }
-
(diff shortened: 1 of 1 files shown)


hirbod:
Just figured there is a const `canUseDOM` - we could even use this

jpudysz:
oh damn, I didn't test it with Next.js. Thanks @hirbod for spotting it out!

hirbod:
@jpudysz no problem. It's nothing breaking, just a little annoyance

mrousavy:
Awesome stuff thanks Hirbod! LGTM

PR #584, opened 2023-09-26
feat: return outcome of set and recrypt.

### LINKED ISSUE #583: Return success state for `set` and `recrypt` calls
The underlying MMKV library returns booleans on these calls wether the `set` or `recrypt` was made successfully or not. It would be nice if these boolean values were to be exposed in the JS world as the application developer may want to do different actions based on the outcome of these method calls. 

https://github.com/Tencent/MMKV/blob/272e5aa1fb57b4ee8cc0ff37e70b7885d99832a7/Core/MMKV_IO.cpp#L1101

https://github.com/Tencent/MMKV/blob/272e5aa1fb57b4ee8cc0ff37e70b7885d99832a7/Core/MMKV_IO.cpp#L557

antondalgren:
This commits adds a boolean return value from the underlying MMKV library to the JS environment for the set and recrypt calls.

closes #583

### DIFF
--- a/android/src/main/cpp/MmkvHostObject.cpp
+++ b/android/src/main/cpp/MmkvHostObject.cpp
@@ -70,18 +70,19 @@ jsi::Value MmkvHostObject::get(jsi::Runtime& runtime, const jsi::PropNameID& pro
                                "MMKV::set: First argument ('key') has to be of type string!");
           }
 
+          bool result = false;
           auto keyName = arguments[0].getString(runtime).utf8(runtime);
 
           if (arguments[1].isBool()) {
             // bool
-            instance->set(arguments[1].getBool(), keyName);
+            result = instance->set(arguments[1].getBool(), keyName);
           } else if (arguments[1].isNumber()) {
             // number
-            instance->set(arguments[1].getNumber(), keyName);
+            result = instance->set(arguments[1].getNumber(), keyName);
           } else if (arguments[1].isString()) {
             // string
             auto stringValue = arguments[1].getString(runtime).utf8(runtime);
-            instance->set(stringValue, keyName);
+            result = instance->set(stringValue, keyName);
           } else if (arguments[1].isObject()) {
             // object
             auto object = arguments[1].asObject(runtime);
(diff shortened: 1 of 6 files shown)


mrousavy:
Is there any case where a `set` could fail? Unless there's some critical memory error or something I feel like this is a bit unnecessary 🤔

antondalgren:
The `set` could fail e.g if there is not enough disk space left.

Follwing is a set of method calls that will show this: 
https://github.com/Tencent/MMKV/blob/272e5aa1fb57b4ee8cc0ff37e70b7885d99832a7/Core/MMKV_IO.cpp#L579 
```cpp 
auto ret = appendDataWithKey(data, key, isDataHolder);
```

in `appendDataWithKey`
https://github.com/Tencent/MMKV/blob/272e5aa1fb57b4ee8cc0ff37e70b7885d99832a7/Core/MMKV_IO.cpp#L797
```cpp
return doAppendDataWithKey(data, keyData, isDataHolder, static_cast<uint32_t>(keyData.length()));
```

in `doAppendDataWithKey`
https://github.com/Tencent/MMKV/blob/272e5aa1fb57b4ee8cc0ff37e70b7885d99832a7/Core/MMKV_IO.cpp#L744-L747
```cpp
    bool hasEnoughSize = ensureMemorySize(size);
    if (!hasEnoughSize || !isFileValid()) {
        return make_pair(false, KeyValueHolder());
    }
```

in `ensureMemorySize`
https://github.com/Tencent/MMKV/blob/272e5aa1fb57b4ee8cc0ff37e70b7885d99832a7/Core/MMKV_IO.cpp#L382-L397
```cpp
return expandAndWriteBack(newSize, std::move(preparedData))
```


in`expandAndWriteBack`
https://github.com/Tencent/MMKV/blob/272e5aa1fb57b4ee8cc0ff37e70b7885d99832a7/Core/MMKV_IO.cpp#L401C21-L429
```cpp
        // if we can't extend size, rollback to old state
        if (!m_file->truncate(fileSize)) {
            return false;
        }

        // check if we fail to make more space
        if (!isFileValid()) {
            MMKVWarning("[%s] file not valid", m_mmapID.c_str());
            return false;
        }
```
As `recrypt` also will end up in the `expandAndWriteBack` function, `recrypt` may also fail.

Without this boolean return value, users of this library may think that the value that was `set` is guaranteed to be persisted, but in fact we can see that if the file can't be expanded, the write will return false and rollback to the old state without notifying the caller.

mrousavy:
Good point. Hm, I'll think about this and come back soon

Anyways, the PR is good & well implemented, thank you so much!

PR #598, opened 2023-11-07
fix: correct behavior of function setState parameter

### LINKED ISSUE #599: "setState" from `useMMKV<Type>` hooks does not behave 100% correctly with function parameter
First off, loving this library! Thanks for all the work on it. I think I've found a bug (and also created a fix PR).

## Bug Summary
Currently, the `setState` value returned from the `useMMKV<Type>` hooks is providing unexpected behavior when the parameter passed to it is a function. Details below.

## Steps to Reproduce
This component demonstrates a minimal reproduction:

```typescript
import React from 'react';
import { Button, Text } from 'react-native';
import { useMMKVNumber } from 'react-native-mmkv';

const MyComponent = () => {
  // We'll track a React state variable,
  const [state, setState] = React.useState(0);
  // AND an MMKV variable.
  const [value, setValue] = useMMKVNumber('my-key');

  return (
    <>
      <Button
        title="Double Increment Me"
        onPress={() => {
          // Increment the state value twice, using the function form of useState.
          setState((current) => current + 1);
          setState((current) => current + 1);

          // Increment the MMKV value twice, using the same function form.
          setValue((current) => (current ?? 0) + 1);
          setValue((current) => (current ?? 0) + 1);
        }}
      />

      <Text>State: {state.toString()}</Text>

      <Text>MMKV: {(value ?? 0).toString()}</Text>
    </>
  );
};
```

If we press the Button in this^ component three times, we will see these values:

```
<after first press>
State: 2
MMKV: 1

<after second press>
State: 4
MMKV: 2

<after third press>
State: 6
MMKV: 3
```

This^ strikes me as incorrect behavior -- it seems like the `useMMKV<type>` hooks are intended to look + feel + behave like `useState`.

## Root Cause
The core problem is [this line](https://github.com/mrousavy/react-native-mmkv/blob/master/src/hooks.ts#L68). In the event of multiple function-based updates, the reference to `valueRef.current` does not change -- it remains as the "stale" "current" value, and does not reflect any just-applied updates.

## Proposed Solution
I've proposed a PR to fix this problem: #598

swain:
Fixes #599.

Please read the linked issue for a detailed description of the issue being fixed here.

## Changes
This change is a one-liner, nice and simple.

I had hoped to write a test to validate the change + prevent future regressions, but I didn't see a test suite. 
I didn't want to make this PR too big, so I opted to add a test suite and a relevant test in a separate PR: #600.

### DIFF
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -59,13 +59,11 @@ function createMMKVHook<
   ): [value: T, setValue: (value: TSetAction) => void] => {
     const mmkv = instance ?? getDefaultInstance();
     const [value, setValue] = useState(() => getter(mmkv, key));
-    const valueRef = useRef<T>(value);
-    valueRef.current = value;
 
     // update value by user set
     const set = useCallback(
       (v: TSetAction) => {
-        const newValue = typeof v === 'function' ? v(valueRef.current) : v;
+        const newValue = typeof v === 'function' ? v(getter(mmkv, key)) : v;
         switch (typeof newValue) {
           case 'number':
           case 'string':


mrousavy:
Thanks for your PR! I read the issue and your diff and yep, it definitely was a bug before.

The reason `valueRef` existed was for performance, as calling `getter(..)` is a C++ JSI call, which is **much** more expensive than just getting a simple JS value (here, a Ref).

This fixes the bug, so let's merge this for now, and I don't think that this performance squeeze is really required here, we're not setting state _that_ often that it really makes a difference.

Thanks man, LGTM!

swain:
@mrousavy Thanks for that additional context, and for the quick review!

Q: do I need to do anything else to get this fix published to NPM? Or, will that happen automatically?

mrousavy:
Will do it soon! Thanks

PR #611, opened 2023-11-23
MultiProcess mode on Android

maxencehenneron:
This PR adds MultiProcess mode support on Android. iOS already supports this mode using AppGroups. 
On android, widgets  run in a different processus and this needs to be enabled to share data between native widgets and the main react-native app.


src/MMKV.ts
```diff
@@ -46,6 +54,20 @@ export interface MMKVConfiguration {
...
+   *
+   * @example
+   * ```ts
+   * const extensionStorage = new MMKV({ mode: MMKVMode.MultiProcess })
+   * ```
+   *
+   * _Notice_: On iOS, this will automatically be set to MMKVMode.MultiProcess if you set an AppGroup in the plist configuration.
+   * More information on AppGroups [here](https://github.com/mrousavy/react-native-mmkv/tree/master#app-groups)
+   */
+  mode?: MMKVMode;
```
mrousavy, line 70: Can you please use TypeScript unions instead of enums here?

So instead of `mode?: MMKVMode` it would be `mode?: 'single-process' | 'multi-process'`

README.md
```diff
@@ -124,6 +125,8 @@ The following values can be configured:
...
 * `encryptionKey`: The MMKV instance's encryption/decryption key. By default, MMKV stores all key-values in plain text on file, relying on iOS's/Android's sandbox to make sure the file is encrypted. Should you worry about information leaking, you can choose to encrypt MMKV. (documentation: [iOS](https://github.com/Tencent/MMKV/wiki/iOS_advance#encryption) / [Android](https://github.com/Tencent/MMKV/wiki/android_advance#encryption))
+* `mode`:  *Android Only*: The MMKV mode. It is set to SingleProcess by default. You can set its value to MultiProcess to support simultaneous read-write access between processus at the cost of performance. This is useful when you want to share data between your react-native app and native extensions such as widgets. _Notice_: On iOS, this will automatically be set to MMKVMode.MultiProcess if you set an AppGroup in the plist configuration (more information on AppGroups [here](https://github.com/mrousavy/react-native-mmkv/tree/master#app-groups))
```
mrousavy, line 128: Is it possible to also use this prop for iOS? I hate having "android only" or "ios only" props... Maybe we can just reconfigure the instance on iOS when this prop is passed?

mrousavy (changes requested):
Hey, thanks for your PR man!

Left some comments

README.md
maxencehenneron: Unfortunately no, I wanted to do this first but on iOS we only have the following constructors: 

Screenshot 2023-11-30 at 9 34 38 AM

As you can see we cannot set both "rootPath" and "mode" at the same time, meaning the only way to set MMKVMode to "multi-process" is to set an AppGroup when initializing MMKV.

maxencehenneron: I could however add support for SingleProcess mode with AppGroups but I'm not sure why would anyone want to do that.

PR #616, opened 2023-12-19
fix: getAllKeys on web

janicduplessis:
Have to make sure to remove the prefix from the keys before returning them.


mrousavy:
ahh, thank you! lgtm

PR #623, opened 2024-01-18
feat: Add `setValue` with function setter for `useMMKVObject`

fahad86:


src/hooks.ts
```diff
@@ -177,24 +177,25 @@ export function useMMKVObject<T>(
   key: string,
   instance?: MMKV
 ): [value: T | undefined, setValue: (value: T | undefined) => void] {
-  const [string, setString] = useMMKVString(key, instance);
+  const [json, setJson] = useMMKVString(key, instance);
+
+  const value = useMemo(() => (json ? JSON.parse(json) : undefined), [json]);
```
mrousavy, line 182: Can you revert the change in `useMemo`? I liked the explicit `if` better, it's more readable.

src/hooks.ts
```diff
@@ -177,24 +177,25 @@ export function useMMKVObject<T>(
...
-    if (string == null) return undefined;
-    return JSON.parse(string) as T;
-  }, [string]);
   const setValue = useCallback(
-    (v: T | undefined) => {
-      if (v == null) {
-        // Clear the Value
-        setString(undefined);
+    (v: (T | undefined) | ((prev: T | undefined) => T | undefined)) => {
+      if (v instanceof Function) {
```
mrousavy, line 186:
```suggestion
      if (typeof v === 'function') {
```

Let's use the `typeof` check instead, we used that everywhere else as well. Theoretically C++ host functions might not be function instances, and I think that check is safer.

src/hooks.ts
```diff
@@ -177,24 +177,25 @@ export function useMMKVObject<T>(
...
+        setJson((currentJson) => {
+          const newValue = v(
+            currentJson ? JSON.parse(currentJson) : undefined,
+          );
```
mrousavy, line 190:
```suggestion
          const currentValue = currentJson != null ? JSON.parse(currentJson) : undefined
          const newValue = v(currentValue);
```

This is also much more readable

src/hooks.ts
```diff
@@ -177,24 +177,25 @@ export function useMMKVObject<T>(
...
+          return JSON.stringify(newValue);
+        });
       } else {
-        // Store the Object as a serialized Value
-        setString(JSON.stringify(v));
+        // Store the Object as a serialized Value or clear the value
+        setJson(v ? JSON.stringify(v) : undefined);
```
mrousavy, line 195:
```suggestion
        const newValue = v != null ? JSON.stringify(v) : undefined
        setJson(newValue);
```

mrousavy (changes requested):
Thanks! Left some styling comments

mrousavy:
Actually you are right, we need to use `instanceof Function` because it might be a non-callable `function` object as well.

PR #643, opened 2024-03-06
Use install_modules_dependencies in podspec

oblador:
Hi Marc 👋 😄 

When using `react-native-mmkv` with latest RN RC (0.74-RC.2), iOS fails to build due to some C++ errors that looked like header issues to me. This PR changes the Podspec to use the `install_modules_dependencies` function if available which solves this issue for me.

### DIFF
--- a/react-native-mmkv.podspec
+++ b/react-native-mmkv.podspec
@@ -28,5 +28,9 @@ Pod::Spec.new do |s|
   ]
 
   s.dependency "MMKV", ">= 1.3.3"
-  s.dependency "React-Core"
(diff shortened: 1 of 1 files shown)


mrousavy:
Hey Joel - nice catch man thank you - didn't know that existed! Do other modules use that function as well? 

If yes and it's meant to be used as a public API, we can merge it!

oblador:
Yes, it’s a public API since 4 minor versions ago or so and I think about half of libraries supporting new arch using it 👍

mrousavy:
let's go then! thanks!

PR #647, opened 2024-03-11
fix: Use `.mock` extension for MMKV mocks

Mister-CK:
The issue I was having that when I was running metro in mock mode. isJest() returned false, so the createMMKV from the regular file was imported, but since I was running in mock mode. That file didn't exist, so createMMKV was undefined. 

Maybe I am missing something, but .mock files are supposed to replace their regular counterparts when running in mock mode. It doesn't make much sense to me to import from both and have a check to determine which to use. Instead, metro  should pick the one that is currently relevant. This is achieved by giving the functions the same name and only importing from the regular one. the .mock one is used when running in mock mode.
For jest to use the mock I have added jest.mock in the unit test. This feels like a more appropriate way to add a mock then changing the imports in the actual code. 

Alternatively, we could add a .mock for that platform checker that always returns true, that is how I initially patched it and a smaller change. 

I am not sure what the .web variant is for, I didn't want to remove it, since it is unrelated, but it does not appear to be used anywhere. 

This is my first PR to this repo, please let me know if I should do something else or in a different way.

### DIFF
--- a/src/MMKV.ts
+++ b/src/MMKV.ts
@@ -1,6 +1,4 @@
 import { createMMKV } from './createMMKV';
-import { createMockMMKV } from './createMMKV.mock';
-import { isJest } from './PlatformChecker';
 
 interface Listener {
   remove: () => void;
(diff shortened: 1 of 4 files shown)


mrousavy:
Thanks for your PR, this looks good to me! I'll do some tests to confirm.

The .mock files will be compiled out I assume?

mrousavy:
Closing as stale: this targets the old package layout and would need to be redone against the current V4/Nitro structure.

PR #653, opened 2024-03-23
chore: add ios privacy manifest

RhyG:
Added a privacy manifest file to comply with Apple's upcoming [privacy manifest requirements](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files).

I based the reason and type off of Apple's docs for [describing use of required reason API](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api) and my understanding of what the library uses, but if it's not quite correct let me know and happy to adjust it.

### DIFF
--- a/example/ios/MmkvExample.xcodeproj/project.pbxproj
+++ b/example/ios/MmkvExample.xcodeproj/project.pbxproj
@@ -394,10 +394,12 @@
 			inputPaths = (
 				"${PODS_ROOT}/Target Support Files/Pods-MmkvExample/Pods-MmkvExample-resources.sh",
 				"${PODS_CONFIGURATION_BUILD_DIR}/React-Core/AccessibilityResources.bundle",
+				"${PODS_ROOT}/../../../ios/PrivacyInfo.xcprivacy",
 			);
 			name = "[CP] Copy Pods Resources";
 			outputPaths = (
 				"${TARGET_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}/AccessibilityResources.bundle",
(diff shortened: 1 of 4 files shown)


jbrodriguez:
the underlying library maintainer is removing uses of the apis that trigger the warning (he doesn't seem willing to include a privacy manifest in his library)

last change was made [2 days ago](https://github.com/Tencent/MMKV/issues/1220), i'm not sure if he cleaned up all instances of "offending" APIs, so it's probably a good idea to still declare the manifest here 🤷‍♂️ ? 

any thoughts @RhyG  ?

thanks for your work btw 👍

mrousavy:
Hey - thanks for this PR, this is dope!

But I agree with @jbrodriguez - the maintainer of Tencent/MMKV is removing all the APIs that require such privacy manifests, and I think that's a better solution moving forward. See commits like https://github.com/Tencent/MMKV/commit/1feba14d3347050c181ab3e507ac8121d69ae656

calebmackdavenport (approved):

mobileteo:
Any update on the release of this?

cvisserMaxedy:
@mrousavy Could you check this PR? It's relatively simple. The alternative is that the manifest has to be set on the app level, which is a little dirty IMO.

mrousavy:
> @mrousavy Could you check this PR? It's relatively simple. The alternative is that the manifest has to be set on the app level, which is a little dirty IMO.

Again, the real solution to this would be to upgrade the core MMKV library and remove all of the sensitive APIs.  You can already do that by just running pod update yourself.

In V3 I've already fixed this.

RhyG:
In that case are you happy for me to close this @mrousavy?

mrousavy:
yup thx

PR #656, opened 2024-03-25
feat: V3 🫴

mrousavy:
## V3

react-native-mmkv V3 brings a ton of new features! 🥳 

### Bridgeless TurboModule

react-native-mmkv is now a bridgeless C++ TurboModule.

This means, there is no more Java or Objective-C code required for module initialization, everything is handled synchronously on the JS thread.

react-native-mmkv already was a JSI module before, so the performance improvements are only marginal.

But the codebase is now shared across iOS and Android, so the implementation is no longer duplicated! 🥳

### New APIs: `trim()` and `size`

V3 adds two new APIs to react-native-mmkv:

- `trim()`: Trim a MMKV storage unit. This **can** be called after deleting a bunch of keys (as deletions don't resize the storage), or when receiving a memory warning by the OS. You don't need to call this though.
- `size`: Gets the current actual size of the storage unit, in bytes.

Example:

```ts
const storage = new MMKV()
if (storage.size >= 400) {
  storage.trim()
}
```

### New config option: `mode`

A new option has been added to the `MMKV` constructor: `mode`.

`mode` can be one of two values:

- `Mode.SINGLE_PROCESS`: The MMKV instance is only used from a single process (this app).
- `Mode.MULTI_PROCESS`: The MMKV instance may be used from multiple processes, such as app clips, share extensions or background services.

Example:

```ts
const storage = new MMKV({
  id: 'shared-storage',
  mode: Mode.MULTI_PROCESS
})
```

### Buffer improvements

The Buffer APIs have improved.

1. Instead of `Uint8Array`, `getBuffer(...)` and `set(.., buffer)` now use `ArrayBuffer` as a type. If you want to write Uint8's, you can still use `new Uint8Array(arrayBuffer)`, but `ArrayBuffer` is the way to go in most cases. 
2. **Performance improvements**: The buffer APIs got much faster in 3 locations:
    1. `ArrayBuffer` is faster to use than `Uint8Array`/`TypedArray`
    2. `set(key, buffer)` now performs one less copy, which could yield huge performance improvements when working with larger buffers.
    3. `getBuffer(...)` now performs one less copy, which could yield huge performance improvements when working with larger buffers.

### MMKV Core on iOS

Since react-native-mmkv V3 is now a cross-platform C++ TurboModule (or xplat/"CxxTurboModule"), we are also now using MMKV-Core instead of MMKV-iOS for the iOS codebase.

For simple values (such as numbers or booleans) this now uses C++ directly, instead of going through Objective-C/NSObjects. This could potentially lead to another performance gain.

---- 

Note: react-native-mmkv V3 requires react-native 0.74 or higher, and requires the new architecture (TurboModules/CodeGen) to be enabled.

For the sake of simplicity, there will be no backwards compatibility for old architecture (bridge), or RN <0.73.


mrousavy:
Currently stuck with this runtime error in Xcode:

```
dyld[41190]: Symbol not found: _objc_claimAutoreleasedReturnValue
  Referenced from: /private/var/containers/Bundle/Application/9C22A145-D026-444A-99F3-BEEA2E6979FB/MmkvExample.app/Frameworks/hermes.framework/hermes
  Expected in: /usr/lib/libobjc.A.dylib
```

mrousavy:
nvm - it runs on iOS! 🎉 

now to Android...

mym0404:
dropping support old arch is kind of a good positive feedback. 

The author and user should make support/use new arch each other for powerful migration with caution about breaking ecosystem.

mrousavy:
I have no information at all from you, I need native logs, I maybe need you to put a breakpoint in the module initializer function, is this on iOS or Android? etc

mrousavy:
@pigeonmal can you try with beta2?

mrousavy:
Yea this seems like a bug in RN Community CLI. The path is wrong when rebuilding.

mrousavy:
CI is green! 🎉🟢😍

react-native-mmkv V3 is ready 💪 

I am a bit hesitant on merging this to main because it requires the new architecture...

Yunnathai:
> dyld[41190]: Symbol not found: _objc_claimAutoreleasedReturnValue

same here 
Sorry, how to solve the same problem?

PR #658, opened 2024-04-07
chore: update usage docs for react-query example

fobos531:
Hello!

I saw this PR got merged: https://github.com/mrousavy/react-native-mmkv/pull/642

After some post-merge discussions in that PR, we came to the conclusion that it might be good to revert the instructions to how they were before, to be more in-line with usage within React.


mrousavy:
Makes sense! Thanks

PR #666, opened 2024-04-24
feat: add automocking in vitest

dcdm3g:
`react-native-mmkv` already has automocking for Jest, but not for Vitest.

This commit adds automocking for Vitest test environments just by checking if the `VITEST_WORKER_ID` environment variable is present.

This new functionality was also documented in the README, in the "Testing with Jest or Vitest" section, formerly the "Testing with Jest" section.

This way, `MMKV` will be mocked in both Jest and Vitest test environments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,9 +191,9 @@ const buffer = storage.getBuffer('someToken')
 console.log(buffer) // [1, 100, 255]
 ```
 
-## Testing with Jest
+## Testing with Jest or Vitest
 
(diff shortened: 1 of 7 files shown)


mrousavy:
Interesting, thanks for your PR!

Changes look good, but do people actually use Vitest in RN apps?

dcdm3g:
> Changes look good, but do people actually use Vitest in RN apps?
[This package](https://github.com/sheremet-va/vitest-react-native) was created to integrate React Native and Vitest, so I am using them together.

About the error during CI, this appears to be an error in the integration between Eslint and Prettier. I will fix this.

dcdm3g:
Oops! I'll see what went wrong and fix it.

mrousavy:
Thank you!

PR #669, opened 2024-04-30
chore: Move everything into `package/`

mrousavy:
Moves the entire MMKV codebase into a package/ subfolder


Babur171:
Getting issues in my redux-persist how to fix it?

**TypeError: storage.getItem is not a function (it is undefined), js engine: hermes**
```
import { Storage } from 'redux-persist'
import { MMKV } from "react-native-mmkv"

const storage = new MMKV()

export const reduxStorage: Storage = {
  setItem: (key, value) => {
    storage.set(key, value)
    return Promise.resolve(true)
  },
  getItem: (key) => {
    const value = storage.getString(key)
    return Promise.resolve(value)
  },
  removeItem: (key) => {
    storage.delete(key)
    return Promise.resolve()
  },
}
```

Issue #691, opened 2024-06-10
Error : TypeError: Cannot read property 'MULTI_PROCESS' of undefined


ashishmangukiya:
Facing issue after enabling bridgeless mode. 

    "react-native-mmkv": "3.0.0-beta.5",
    "react-native": "0.74.2",
    "react": "18.2.0",

// index.js
import { MMKV, Mode } from 'react-native-mmkv';
const storage = new MMKV({ mode: Mode.MULTI_PROCESS });
storage.set("test", "test")

Error : 
TypeError: Cannot read property 'MULTI_PROCESS' of undefined

mrousavy:
created a fix: https://github.com/mrousavy/react-native-mmkv/pull/692

PR #692, opened 2024-06-10
fix: Export `Mode` enum as value

mrousavy:
Previously it was only exported as a type, causing it to be undefined in release.

- Fixes https://github.com/mrousavy/react-native-mmkv/issues/691

### DIFF
--- a/package/src/index.ts
+++ b/package/src/index.ts
@@ -1,4 +1,4 @@
 export * from './MMKV';
 export * from './hooks';
 
-export type { Mode, Configuration } from './NativeMmkv';
+export { Mode, Configuration } from './NativeMmkv';


ashishmangukiya:
@mrousavy  when the issue will be released? 
Currently using 
"react-native-mmkv": "3.0.0-beta.5",
"react-native": "0.74.2",
"react": "18.2.0",
and getting error like 
Invariant Violation: "MyAppName" has not been registered. This can happen if:
* Metro (the local dev server) is run from the wrong folder. Check if Metro is running, stop it and restart it in the current project.
* A module failed to load due to an error and `AppRegistry.registerComponent` wasn't called.

working fine if I use new MMKV()
if I use like ->  new MMKV({ mode: 1 }) then getting above error.

@mrousavy  can you please help me ?

mrousavy:
Dude I told you before in the other issue you created - please stop pinging me

PR #697, opened 2024-06-22
chore: remove leftover code

itsramiel:
## Summary
Since the mmkv was migrated to a pure c++ turbo module, there is no longer need for the adapter which anyway includes useless code


mrousavy:
Thank you!

Issue #615, opened 2023-12-12
The enableKeyExpire function and the expireDuration parameter are missing


davistsin:
MMKV provide function (Global Expiration)
```
// expire in a day
mmkv.enableAutoKeyExpire(MMKV.ExpireInDay); // MMKV.ExpireInDay = 24 * 60 * 60
// enable auto key expiration without global duration
mmkv.enableAutoKeyExpire(MMKV.ExpireNever); // MMKV.ExpireNever = 0
```
And support `set` function parameter uint32_t expireDuration
```c++
bool MMKV::set(int32_t value, MMKVKey_t key, uint32_t expireDuration) {
}
```
But not seen in this library.

I hope `react-native-mmkv` can support those feature.

davistsin:
Also wanna add trim() function. 
```
The totalSize() of an MMKV instance won't reduce after deleting key-values, call this method after lots of deleting if you care about disk usage. Note that clearAll() has a similar effect.
```

mrousavy:
> Also wanna add trim() function.
https://github.com/mrousavy/react-native-mmkv/pull/461

lohenyumnam:
Any Plan to add this feature?? @mrousavy

mrousavy:
Not in my free time, no

zaferatli:
i could help about this, @mrousavy this enhancement is it ok for you?

mrousavy:
Hey - I dont think this is a necessary addition to the library and might just overcomplicate things. It's better to handle expiration yourself imo.

PR #704, opened 2024-07-16
feat: expireDuration feature added

zaferatli:
related enhancement  https://github.com/mrousavy/react-native-mmkv/issues/615

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,6 +225,20 @@ if (size >= 4096) {
 }
 ```
 
+### EnableAutoKeyExpire
+
+```js
+// Enable auto expire for keys
+/*
+ *  use `enableAutoKeyExpire` for enable auto key expire (in seconds)
+ *  -1 => Disable auto key expire
+ *  0 => Never expire (Recommended)
+ *
+ */
+storage.enableAutoKeyExpire = 0; // Keys never expire
+storage.set("user.age", 21, 10); // User.age key expire in 10 second, other keys never expire
+```
+
 ## Testing with Jest or Vitest
 
 A mocked MMKV instance is automatically used when testing with Jest or Vitest, so you will be able to use `new MMKV()` as per normal in your tests. Refer to [package/example/test/MMKV.test.ts](package/example/test/MMKV.test.ts) for an example using Jest.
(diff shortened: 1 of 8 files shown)


mrousavy:
Hey - thank you for your PR, this is great!

One small change - can you change `enableAutoKeyExpiration` to be a getter + setter boolean property, instead of a function? Thanks!

And also, make sure to run lint (`yarn check-all`) to format the C++ code so the CI is happy.

zaferatli:
i change `enableAutoKeyExpiration ` function to getter + setter without boolean property because this function takes number parameter.

i use -1 pointer for `disableAutoKeyExpiration`, its came to me more easy-to-use but if want to change i could

mrousavy:
ohhh - I'm sorry I misunderstood that then. I thought it was a simple boolean!

Hmm, let me think about this, not sure if I want this to be a prop or a function now..

I'll get back to you! :)

zaferatli:
have you had a chance to review it? i think we can use function its would be more easy-to-use

mrousavy:
Hey - so two thoughts:

1. I don't like the way every `set` function now has an extra branch. To me, it looks much more complex than before. I want to avoid that. If you take a look at what `MMKV` itself does internally, is it just passes the currently configured key duration as a default value. So I think we can also do that if we can access that maybe? 
2. If it is a property, it needs to be `number | undefined`, as in either a number or not set. But I think a function makes more sense.

I'm unfortunately still not happy about both solutions, I think this adds a lot of unnecessary complexity and I don't really get what you would need that for? Why do you need your keys to expire? Can you give me some usage examples?

zaferatli:
i just saw related issue, and i've wanted to do contributed it and if u want to put behind until we need its ok for me

mrousavy:
Thank you - I really appreciate the contribution - but I want to avoid merging something that overcomplicates the codebase, because that makes it harder to contribute or improve the project in the future.

erie-e9:
Please merge it :)

techied:
Any hope for this or #818 ? This feature would be very useful for us.

> Why do you need your keys to expire? Can you give me some usage examples?

To answer your question, it's perfect for cache purposes.

Without this, in order to prevent the cache from growing out of control, we have to store dates with each item and read out all of the values in order to do a vacuum job of sorts.
 
The other PR is even simpler and smaller, if you're concerned about maintainability

zaferatli:
stale

celestial-rose:
Hey! Thank you for such an amazing package. I finally have the usage for it after years of longing this moment, and I see that this TTL hasn't merged yet; any update on that matter? What's blocking thing ? Would love to contribute

PS : I saw this just came out https://github.com/janis-commerce/app-storage

PR #703, opened 2024-07-16
feat: If AppGroup defined in plist apply group dir.

calism2:
This commit adds functionality to check if an AppGroup is defined in the info.plist. If it is, the group directory is applied, allowing for shared storage between the main app and its extensions.

Closes #702

Issue #702, opened 2024-07-16
Support for App Groups in MMKV 3.0.0-beta.6


calism2:
Hello,

I'm currently developing an iOS share extension and I believe that the MMKV library would be a great fit for my project. MMKV's compatibility with both React Native versions and its support for app extensions make it an ideal choice. However, I noticed that although the documentation mentions support for app groups via info.plist, this feature seems to be missing in release 3.0.0-beta.6.

As someone who is not an iOS developer, I have managed to implement the basic functionality for sharing app groups. However, having this feature natively supported in MMKV would significantly simplify development and improve the overall experience.

Could you please consider adding support for app groups in the upcoming releases? Thank you.

//  MmkvHostObject.cpp
```
#ifdef __APPLE__
#include <Foundation/Foundation.h>
  NSString* appGroup = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"AppGroup"];
    if( appGroup ){
        NSString* groupDir = [[NSFileManager defaultManager]
                              containerURLForSecurityApplicationGroupIdentifier:appGroup]
            .path;
            const char* groupDirCStr = [groupDir UTF8String];
            std::string groupDirStr(groupDirCStr);
            
            instance = MMKV::mmkvWithID(config.id, mode, encryptionKeyPtr, &groupDirStr);

    }else{
        instance = MMKV::mmkvWithID(config.id, mode, encryptionKeyPtr, pathPtr);
    }
```

mrousavy:
Hey - please create a PR and I'll gladly review & merge changes if everything works! Thanks

Issue #650, opened 2024-03-19
Crash reports after switching to use AppGroups due to 0xdead10cc


mattsellars123:
Hello,

We are using MMKV in a react native app and were hoping to share our mmkv stores using AppGroups. After configuring things everything seemed to work fine for a while then we started getting crash reports with: 

`Termination Reason: RUNNINGBOARD 0xdead10cc`

We do not yet have any extensions writing to this store. I was wondering if there is some other configuration we need to add so that iOS isn't killing our process when the user happens to background it during file writes. 

Any insight is very appreciated.

mrousavy:
I need full crash logs.

mattsellars123:
Here you go. Thanks for taking a look. Generally what I've seen from other file uses for iOS specifics are listening on backgrounding event from App delegate and using the `UIApplication.shared.beginBackgroundTask` but this is very iOS/UIKit specific in this case and from the react side there are many mmkv uses we would need to ensure are sync'd out to the filesystem and released/closed.

[2024-03-19_11-30-10.6591_-0700-2bd285d62f19f77b80ec7dd30686dfc7c6f74e90.crash.log](https://github.com/mrousavy/react-native-mmkv/files/14704732/2024-03-19_11-30-10.6591_-0700-2bd285d62f19f77b80ec7dd30686dfc7c6f74e90.crash.log)

Issue #627, opened 2024-02-07
App group enhancement introduced a breaking change (2.10.0)


JP250552:
The change introducing app groups moved the storage location when the app groups property is set which is a breaking change - https://github.com/mrousavy/react-native-mmkv/pull/402/files#diff-b667beab6bbb6d80aab761c6781a9a7fea5812e5814d98666080a2ec6edb38dfR46

Upgrading to 2.10.0 causes our users to lose stored data since we had an app group previously defined for other reasons.

It would probably be good to revert and release that change as v3 and also provide a way to override that storage location for backwards comparability.

mrousavy:
Oh, whoops. Hmm I don't think reverting that change makes sense as that was more than a year ago and most people are already on that version by then, meaning they'll lose any data unless they upgrade to v3.

I think you could easily migrate by moving the mmkv file from the normal directory to the appgroup directory.

JP250552:
That's a fair point. Would you be opposed to us adding a property to allow using the default location?

PR #705, opened 2024-07-22
feat: Support AppGroups

mrousavy:
Adds support for using AppGroups.

This stores MMKV instances in a shared AppGroup directory - it is mostly automatic.

To do this simply add your `AppGroup` to `Info.plist`, and don't pass a `path` to `new MMKV(...)` - MMKV will then automatically use the `AppGroup` path instead.

### Related issues

- Fixes https://github.com/mrousavy/react-native-mmkv/issues/622
- Fixes https://github.com/mrousavy/react-native-mmkv/issues/627
- Fixes https://github.com/mrousavy/react-native-mmkv/issues/650
- Resolves https://github.com/mrousavy/react-native-mmkv/issues/702 

### Related PRs

- Closes https://github.com/mrousavy/react-native-mmkv/pull/703

Sorry @calism2 - your PR looked good, but I felt like moving the iOS-specific code into `MmkvPlatformContext` made more sense instead of cramming it into the shared cpp constructor.

### DIFF
--- a/package/.eslintrc.js
+++ b/package/.eslintrc.js
@@ -4,6 +4,5 @@ module.exports = {
   ignorePatterns: ['scripts', 'lib'],
   rules: {
     'prettier/prettier': ['warn'],
-    '@typescript-eslint/consistent-type-imports': 'warn',
   },
 };
--- a/package/android/src/main/java/com/mrousavy/mmkv/MmkvPlatformContextModule.java
+++ b/package/android/src/main/java/com/mrousavy/mmkv/MmkvPlatformContextModule.java
@@ -1,5 +1,7 @@
 package com.mrousavy.mmkv;
 
+import androidx.annotation.Nullable;
+
 import com.facebook.react.bridge.ReactApplicationContext;
 
 public class MmkvPlatformContextModule extends NativeMmkvPlatformContextSpec {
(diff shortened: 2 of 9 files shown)


calism2:
It's not a problem that my original PR wasn't merged. The new approach is more readable and has better error handling. I tested both scenarios, and they work perfectly. I've added screenshots as well. Thank you for the quick PR merge.

With app group setup.
Screenshot 2024-07-22 at 21 00 53

Without app group setup.
Screenshot 2024-07-22 at 21 00 53

mrousavy:
Thank you for testing @calism2!

But when do you see this error? Did you set `AppGroup` to something that is not a valid path?

calism2:
Yes, the AppGroup needs to be added through Capabilities (you can add it for all, debug, or release), similar to Apple Pay. If you disable or remove the capability, the path won't be discoverable. As shown in the image, this setup only works for the debug version; you'll encounter an error in the release version. I tested this by removing the AppGroup capability from debug version.

Screenshot 2024-07-22 at 21 35 04

mrousavy:
gotcha - great!

coopermaruyama:
nice! thanks for the contributions 🙂

fivecar:
I'm currently running 2.x because I can't quite migrate to the new TurboModules architecture yet. When I do make the migration to 3.x, what happens to existing keys if I don't specify a `path` (but I do have an App Group)? Do I need to do a manual migration, given 3.x now defaults to storing in the App Group?

pfcodes:
Is there any way to use AppGroups in 2.x? Can't migrate to the new architecture any time soon

PR #711, opened 2024-08-02
Reduce log overhead when logging disabled

### LINKED ISSUE #710: MmkvLogger creates log strings even if logging is disabled
The string_format method seems like an anti-pattern...   Is is possible to move all formatting to platform log implementations, which skip the string format overhead when not in use?

isaacl:
Avoid a string format unless logging is enabled.

Fixes #710.

### DIFF
--- a/package/android/src/main/cpp/AndroidLogger.cpp
+++ b/package/android/src/main/cpp/AndroidLogger.cpp
@@ -8,9 +8,9 @@
 #include "MmkvLogger.h"
 #include <android/log.h>
 
-void MmkvLogger::log(const std::string& tag, const std::string& message) {
+void MmkvLogger::log(const std::string& tag, const std::string& formatString, Args... args) {
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wformat-security"
-  __android_log_print(ANDROID_LOG_INFO, tag.c_str(), message.c_str());
+  __android_log_print(ANDROID_LOG_INFO, tag.c_str(), formatString.c_str(), args);
 #pragma clang diagnostic pop
 }
--- a/package/cpp/MmkvLogger.h
+++ b/package/cpp/MmkvLogger.h
@@ -11,25 +11,12 @@ class MmkvLogger {
 private:
   MmkvLogger() = delete;
 
-private:
-  template <typename... Args>
-  static std::string string_format(const std::string& format, Args... args) {
-    int size_s = std::snprintf(nullptr, 0, format.c_str(), args...) + 1; // Extra space for '\0'
-    if (size_s <= 0) {
-      throw std::runtime_error("Failed to format string!");
-    }
-    auto size = static_cast<size_t>(size_s);
-    std::unique_ptr<char[]> buf(new char[size]);
-    std::snprintf(buf.get(), size, format.c_str(), args...);
-    return std::string(buf.get(), buf.get() + size - 1); // We don't want the '\0' inside
-  }
 
 public:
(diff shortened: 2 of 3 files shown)


isaacl (commented):
Note: this doesn't compile, but it's reasonably close to a working impl

package/cpp/MmkvLogger.h
```diff
@@ -11,25 +11,14 @@ class MmkvLogger {
...
 
 public:
-  static void log(const std::string& tag, const std::string& message);
+  static void log(const std::string& tag, const std::string& formatString, Args... args);
 
   template <typename... Args>
   inline static void log(const std::string& tag, const std::string& formatString, Args&&... args) {
-    std::string formattedString = string_format(formatString, std::forward<Args>(args)...);
-    log(tag, formattedString);
+#ifdef DEBUG
```
isaacl, line 20: this is redundant w/ the directive in iOS.  It might be more correct to just guard iOS which has no concept of log levels, and leave Android based on log level.   Or you could choose to have the directive here to have more similar behavior between iOS and Android

package/cpp/MmkvLogger.h
```diff
@@ -11,25 +11,14 @@ class MmkvLogger {
...
-    }
-    auto size = static_cast<size_t>(size_s);
-    std::unique_ptr<char[]> buf(new char[size]);
-    std::snprintf(buf.get(), size, format.c_str(), args...);
-    return std::string(buf.get(), buf.get() + size - 1); // We don't want the '\0' inside
-  }
 
 public:
-  static void log(const std::string& tag, const std::string& message);
+  static void log(const std::string& tag, const std::string& formatString, Args... args);
```
mrousavy, line 16: This is not a template, you cannot use `Args...` here. This is why the C++ build is now failing.

package/ios/AppleLogger.mm
```diff
@@ -8,9 +8,11 @@
 #import "MmkvLogger.h"
 #import <Foundation/Foundation.h>
 
-void MmkvLogger::log(const std::string& tag, const std::string& message) {
+void MmkvLogger::log(const std::string& tag, const std::string& formatString, Args... args) {
```
mrousavy, line 11: same here, not a template.

we can't even template this because our implementation is in two differnt spots (ios and android). so we might not be able to support `Args...`.

package/android/src/main/cpp/AndroidLogger.cpp
```diff
@@ -8,9 +8,9 @@
 #include "MmkvLogger.h"
 #include <android/log.h>
 
-void MmkvLogger::log(const std::string& tag, const std::string& message) {
+void MmkvLogger::log(const std::string& tag, const std::string& formatString, Args... args) {
```
mrousavy, line 11: same here, not a template.

we can't even template this because our implementation is in two differnt spots (ios and android). so we might not be able to support `Args...`.

mrousavy (changes requested):
thanks for your PR - but C++ build now fails (see other comments), and it's not linted properly (run `yarn lint-cpp`)

package/cpp/MmkvLogger.h
isaacl: yes, I understand, but I ended up not using this project for my work, so I don't have time to fully flesh out this patch.  I was intending to simply illustrate some ways of avoiding the logging overhead and/or avoiding NSLog in release builds (which is discouraged)

mrousavy:
Closing as stale: this branch has requested changes and no longer applies cleanly to the current native layout.

PR #719, opened 2024-08-19
fix: Update cpp code to support React Native 0.75.1

davebray131:
Updates cpp code to rename some variables to meet new CodeGen specifications


mrousavy:
quite a breaking change 😅

mrousavy:
please run clang format.

mrousavy:
Both iOS and Android builds failed now, so this PR is definitely not backwards compatible. I don't want to introduce a lot of new code with `#ifdef` and stuff, so maybe I'll just wait a bit before I merge & release this.

People will complain either way

davebray131:
I think that it is perfectly fine to do that.  I am content using the patch I created that was posted in the original issue, and I think that will work well for anyone that needs support for 75.1+.  Once that version gains more traction these changes can be integrated into the main branch.  Thanks for taking the time to look at it.

alon7:
Awesome work, was experiencing the same thing

PR #739, opened 2024-09-08
fix "Identifier 'module' has already been declared" on web v3.0.2

arasrezaei:
when using react compiler with expo, i got this error on web:
`Identifier 'module' has already been declared`
simply renaming that `module` to `Module` fixed the issue.


mrousavy:
Ah interesting! Instead of `Module`, can you please rename it to `turboModule`? Variables shouldn't start with uppercase letters.

mrousavy:
Awesome, thank you!!

PR #742, opened 2024-09-12
chore(docs): update README

stephane-r:
I tried a lot to understand why react-native-mmkv did not work on my RN version in 0.74.x, before finding information on an issue which is closed.

Maybe the information is important until version 0.76 ?


mrousavy:
Good point!

PR #743, opened 2024-09-18
feat: allow codesplitting mock file

CorentinAndre:
## Description

Hey 👋 first of all, thanks for your lib Mark :)

Here's a small PR where the mock and native file module are required inline. Otherwise, if run in a test environment, it will parse the native file and crash because it's importing react-native.


mrousavy:
bump @CorentinAndre

Issue #713, opened 2024-08-05
V3 Beta drops support for tvOS?


Duell10111:
I saw that the podspec of the new v3 versions does not contain the tvOS platform anymore.
https://github.com/mrousavy/react-native-mmkv/blob/8b5e9d009c0e7da47d5bf4f3705b01f570f09c26/package/react-native-mmkv.podspec#L17
Is this intended, as this would break my current Expo TVOS app. :D

mrousavy:
hans stop talking bullshit

tvOS did work, I think this is a mistake - can you make a PR to add it again and verify that this works? thanks! :)

mrousavy:
😂

mrousavy:
@Duell10111 can you add a PR to make tvOS work again?

mrousavy:
I think this has been fixed in [react-native-mmkv 4.x.x-beta](https://github.com/mrousavy/react-native-mmkv/releases/tag/v4.0.0-beta.0). It should work on tvOS again

PR #745, opened 2024-09-23
fix: enable macOS and tvOS in podspec

thejustinwalsh:
Yolo'd this to unblock my react-native-macos app, which I am migrating to the new architecture.

With the new architecture enabled, I tested this change for macOS by running it with [react-native-test-app](https://github.com/microsoft/react-native-test-app).

I restored the podspec to the [previous code](https://github.com/mrousavy/react-native-mmkv/commit/3d0a289a7bf2ff1d67920890b9a740a435c5d39b) that was here in the history using git blame. I did not test tvOS support against this change, but I saw that #713 was open and asked for this change as well.

What else would you like to see in this PR? Examples, tests, etc.?
Also, it is unclear to me what the min versions should be in the podspec for tvOS and macOS.

### DIFF
--- a/package/react-native-mmkv.podspec
+++ b/package/react-native-mmkv.podspec
@@ -14,7 +14,7 @@ Pod::Spec.new do |s|
   s.license      = package["license"]
   s.authors      = package["author"]
 
-  s.platforms    = { :ios => min_ios_version_supported }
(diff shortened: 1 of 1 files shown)


mrousavy:
YOLO LGTM - thanks for your PR!!! :)

PR #777, opened 2024-12-08
Use subscribe instead of clearAll

rklomp:
See https://jotai.org/docs/utilities/storage#parameters
The 4th parameter should be a subscribe function. The clearAll function does not make sense here.

### DIFF
--- a/docs/WRAPPER_JOTAI.md
+++ b/docs/WRAPPER_JOTAI.md
@@ -21,8 +21,18 @@ function removeItem(key: string): void {
   storage.delete(key);
 }
 
-function clearAll(): void {
-  storage.clearAll();
+const subscribe = (key: string, callback: (value: string | null) => void) => {
+  const listener = (changedKey: string) => {
+    if (changedKey === key) {
+      callback(getItem(key))
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


rklomp:
It is also a good idea to have `getOnInit` set to `true` to load values from MMKV immediately instead of having a rerender because the data was loaded. Do we want to add this to the documentation as well?

```diff
const persistAtom = <T>(key: string, initialValue: T) =>
  atomWithStorage<T>(
    key,
    initialValue,
    createJSONStorage<T>(() => ({
      getItem,
      setItem,
      removeItem,
      subscribe
    })),
+   { getOnInit: true }
  )
```

mrousavy:
yea let add that as well

rklomp:
> yea let add that as well

Done :)

mrousavy:
maybe that was called for some people and their DB was deleted? Could that be the case? I've heard reports from people saying that happened, not sure if it was Jotai.

But thanks for the PR!

redkean:
oh, it's nice that `{ getOnInit: true }` is added here. i was thinking on making a PR for this because I encountered an issue on atomWithStorage that returns the initial value instead of the saved value on MMKV.

mrousavy:

PR #800, opened 2025-02-16
feat: Upgrade to react-native 77

mrousavy:
Upgrades the package & example to react-native 0.77.

This allows for a new zero-copy string API which I can use to make MMKV even faster


mrousavy:

PR #801, opened 2025-02-16
perf: Make all methods a bit faster by skipping double JSI type-check

mrousavy:
Makes all MMKV methods a bit faster by skipping the type-check after we already checked the type. Replace `asString` -> `getString`, we already checked `isString` before.


mrousavy:

PR #803, opened 2025-02-16
fix: Add `readOnly` to `new(..)` types

mrousavy:
The `readOnly` configuration property was supported on native, but was missing from the types when calling `new MMKV(...)`. This PR adds it there.

If you don't need to write to the MMKV database, creating it in `readOnly` mode is more efficient:

```ts
const instance = new MMKV({ id: 'storage', readOnly: true })
```


mrousavy:

Issue #812, opened 2025-02-19
Data Loss Migrating from V2 -> V3 with app group set on iOS


JacquesLeupin:
When migrating from V2 -> V3 in an Expo app on iOS, application data stored through `react-native-mmkv` is lost. The data loss occurs as part of an upgrade we're attempting from old arch -> new arch, react native version 0.74.5 -> 0.76.7, and Expo SDK 51 -> Expo SDK 52. Attempting `react-native-mmkv` upgrade from version `2.12.2` to version `3.2.0`

Our app contains an AppGroup for an iOS share extension.  

We have a simple file which initializes the mmkv store without any configuration:

```typescript
import { MMKV } from 'react-native-mmkv';

export const mmkvStorage = new MMKV();
```

I've tried many times to alter the mmkv config while upgrading to not lose the data without success (altering the config when initializing, maintaining a stable MMKV version/submodule from a forked version of the package, etc). Nothing's worked thus far.

Here are the native logs when `react-native-mmkv` initializes as version `2.12.2`, part of Expo 51, RN 0.74.5:

```c++
Installing global.mmkvCreateNewInstance...
[I] <MMKV.cpp:167::initialize> version v2.0.2, page size 16384, arch arm64-v8a
[I] <MMKV.cpp:219::initializeMMKV> Apple Device: 3, version: 17, mlock enabled: 0
[I] <MMKV.cpp:228::initializeMMKV> root dir: /var/mobile/Containers/Data/Application/D6E26C10-0D3C-4D4D-862B-D7284BD0B57C/Documents/mmkv
[I] <libMMKV.mm:132::+[MMKV initializeMMKV:groupDir:logLevel:]> groupDir: /private/var/mobile/Containers/Shared/AppGroup/846A9DAC-AB6F-40F2-82D4-A8A3C3A0999E/mmkv
Installed global.mmkvCreateNewInstance!
[D] <libMMKV.mm:927::+[MMKV mmapKeyWithMMapID:rootPath:]> mmapKey: 1df96a782b76e1e46f48a86659296c95
[I] <MMKV.cpp:255::mmkvWithID> prepare to load mmkv.default (id 1df96a782b76e1e46f48a86659296c95) from rootPath /private/var/mobile/Containers/Shared/AppGroup/846A9DAC-AB6F-40F2-82D4-A8A3C3A0999E/mmkv
[...]

PR #813, opened 2025-02-22
fix: Append mmkv to app group default path to mimick MMKV library

JacquesLeupin:
This is a proposed fix for #812.

When upgrading from v2 to v3 of the library *with an app group on iOS*, mmkv data is lost.

The data loss occurs when we pass the app group path when initializing the native MMKV instance without appending `/mmkv` to that path (in doing so, we diverge from the default behavior of the library as outlined here https://github.com/Tencent/MMKV/blob/64c3d03951dd1c784bb2eeb5561b1899f56d1155/iOS/MMKV/MMKV/MMKV.h#L56).

Not sure if this is the best way to achieve the fix (or how to test it), so suggestions/enhancements are welcome.

### DIFF
--- a/package/src/createMMKV.ts
+++ b/package/src/createMMKV.ts
@@ -14,7 +14,8 @@ export const createMMKV = (config: Configuration): NativeMMKV => {
           getMMKVPlatformContextTurboModule().getAppGroupDirectory();
         if (appGroupDirectory != null) {
           // If we have an `AppGroup` in Info.plist, use that as a path.
-          config.path = appGroupDirectory;
+          // We append mmkv to the path to conform to the default behavior of the MMKV library.
+          config.path = appGroupDirectory + '/mmkv';
         }
       } catch (e) {
         // We cannot throw errors here because it is a sync C++ TurboModule func. idk why.


mrousavy:
Thanks for your PR, but this looks like a major breaking change and will cause data loss for everyone using app groups right now..?

JacquesLeupin:
> Thanks for your PR, but this looks like a major breaking change and will cause data loss for everyone using app groups right now..?

Yes, it'd be a breaking change, with the added caveat that it would affect every using app groups *who haven't specified a storage path in their config*.

I'm not quite sure what a better solution would be at this point. Modifying the API to expose the `getAppGroupDirectory` doesn't seem correct.

Without a fix, however, all users using app groups without a path in their config will experience data loss when upgrading from v2 to v3.

JacquesLeupin:
Two backwards-compatible options I could think of to fix this:

- run a full fledged migration when initializing based on the presence of app group shared dir and data in that group
- modify the config to expose an optional boolean property to alter the path

IMO neither option is great for a library that's supposed to be a simple wrapper.

Maybe the `README.md` should state that the behavior of the wrapper differs from native MMKV behavior on iOS with an app group configured and no path set in config?

Given this PR is potentiall a breaking change for existing v3 users, I'll close it down and simply patch the package when upgrading to v3.

mrousavy:
I'd say it's too late to fix this.

V2 is quite old already, most people are on V3.

JacquesLeupin:
>I'd say it's too late to fix this.
>
>V2 is quite old already, most people are on V3

Yes that seems reasonable. In any case, thanks for the library @mrousavy !

PR #818, opened 2025-03-09
Feat :- Add Support for auto key expire

riteshshukla04:
This PR adds support for automatic key expiration in storage. With this feature, keys will automatically expire after a specified number of seconds if enabled .

Uses MMKV core's enableAutoKeyExpire

Usage:-

```js
storage.enableAutoKeyExpire(seconds);
```


fobos531:
This is actually a very cool method!

Issue #815, opened 2025-02-27
Support of types keys for storage instance


ipakhomov:
Hey there, thanks for amazing library!

Currently, keys for values in a storage aren't typed, they are just `string`s. 

Having ability to pass an optional generic type when initializing a new mmkv instance could improve developer experience.
This will allow to specify which values are usable with particular storage instance. Example:

```ts

type StorableValues = 'token' | 'userId';

export const storage = new MMKV<StorableValues>({ ... });
...
storage.getString('foo'); // Typescript will report this error
storage.getString('token'); // All good
```
Of course, this can be achieved using a separate enum for keys in the app, but having generic support is more elegant and easier in use.

Please let me know if this feature is desirable, and you'd like me to submit a PR with these changes!

mrousavy:
Closing as per discussion in #826.

Appreciate the time spent on this anyways! ❤️

PR #824, opened 2025-03-31
Support of types keys for storage instance

Zcating:
#815


mrousavy:
Closing as stale: this targets the old package layout and would need to be redone against the current V4/Nitro API.

Issue #808, opened 2025-02-17
How to use react-native-mmkv v3 with custom native code


roni-castro-shipt:
### Description

I currently use MMKV for obtaining an anonymous ID from storage on both Android and iOS in a React Native app (version 0.75.5). The code snippet below initializes MMKV and retrieves a string value:

```objc
#import <MMKV/MMKV.h>

- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
{
   [MMKV initializeMMKV:nil];
   MMKV *mmkv = [MMKV defaultMMKV];
   NSString *userInfo = [mmkv getStringForKey:storageKey];
   // do some stuff with that userInfo, and this works as expected when using the Xcode debugger
  ...
}
```

- I've set `ENV['RCT_NEW_ARCH_ENABLED'] = '1'`
- I'm on React Native version 0.75.5.
- I don't have `use_frameworks` in my `ios/Podfile`.

After updating MMKV to version from 2.11.0 to `3.2.0`, I encountered an error indicating that `#import <MMKV/MMKV.h>` could not be found, as it seems the native peer dependency for MMKV was removed in this latest version.

![Image](https://github.com/user-attachments/assets/249b104e-0e78-4eff-a635-33bb757da2e2)

To fix this, I tried manually adding the MMKV pod to my `Podfile` like so:

```ruby
pod 'MMKV', :modular_headers => true
```

However, after doing this, I'm encountering a crash related to buffers (I'm unsure if this crash is caused by MMKV or another library that depends on MMKV).

![Image](https://github.com/user-attachments/assets/73730adc-a868-4cac-9f5f-2d49ffcff16b)

---

### Question:

For this version of the library (MMKV 3.2.0), why do I need to manually add the MMKV pod instead of relying on autolinking?

quangcao-offerup:
same issue for me when upgrading from `2.12.2` to `3.2.0`, . Adding that back to pod file, it seem to show previous version installed. might be a better way. 

![Image](https://github.com/user-attachments/assets/a6508bad-c106-4a1f-8248-d0725706a916)

gfrancischini:
Hey @mrousavy  do you think that using the newest version from mmkv 2.1.0 (Supports using MMKV directly in C++ code.) would be possible to share the lib between native and react native?

https://github.com/Tencent/MMKV/releases/tag/v2.1.0

If so, can we talk about sponsoring your work to do so?

Thanks

[...]

PR #828, opened 2025-04-08
feat: Use official MMKV library from Gradle & Pod instead of prebuilding it

mrousavy:
- Implements https://github.com/mrousavy/react-native-mmkv/issues/808


mrousavy:
closed in favor of brad's PR

PR #831, opened 2025-05-13
fix: funding-octocat.svg link on README.md

renanmav:
Link was broken


renanmav:
It was a browser issue

PR #841, opened 2025-07-13
Set C++ standard to C++20 in CMake configuration

riteshshukla04:
image

I am facing this issue in latest RN(0.81-rc0). As per error RN uses` requires` which is a c++ 20 feature. So build is failing.


mrousavy:
cool! thx

sharifhh:
can you please publish a beta or a new version with this fix? 🙏 
@mrousavy

dppo:
@mrousavy 
The same issue appears in React Native 0.81.
I noticed that v4.0.0 was rewritten to a Nitro Module.
Is it possible to release a new version on top of v3?

PR #846, opened 2025-08-10
feat: use pod and prefab

boorad:
- updated `feat/pod` branch to `main`
- created `feat/pod-boorad` branch
- fixed `ios` and `android` builds to work with pod and prefab
- fixed dark mode bugs
- added a ton of local build shit that you might not want.  But it helped me work with the project in VSCode-like IDEs
- changed `example` app more than I wanted, with `prefab: true` and dep on `mmkv-shared`

Take a look... this works, but could probably be improved.


mrousavy:
@boorad can we target `main` instead of my branch? This could be merged standalone right?

package/cpp/MmkvTypes.h
```diff
@@ -0,0 +1,63 @@
...
+constexpr auto MMKVLogInfoValue = ::mmkv::MMKVLogInfo;
+constexpr auto MMKVLogWarningValue = ::mmkv::MMKVLogWarning;
+constexpr auto MMKVLogErrorValue = ::mmkv::MMKVLogError;
+constexpr auto MMKVLogNoneValue = ::mmkv::MMKVLogNone;
+
+// Mode constants
+constexpr auto MMKVSingleProcessValue = ::mmkv::MMKV_SINGLE_PROCESS;
+constexpr auto MMKVMultiProcessValue = ::mmkv::MMKV_MULTI_PROCESS;
+
+#endif
```
mrousavy, line 63: Couldn't we theoretically just do `using namespace mmkv` then it works the same on all platforms?

I don't like that we have to use different names (e.g. `MMKVType` instead of `MMKV`) in our codebase. I guess it makes sense and is not ambiguous, but I'm just thinking if there's alternatives.

package/cpp/NativeMmkvModule.h
```diff
@@ -7,9 +7,11 @@
 
 #pragma once
 
-#if __has_include(<React-Codegen/RNMmkvSpecJSI.h>)
+#include <string>
```
mrousavy, line 10: lol does codegen not include that?

package/example/android/app/build.gradle
```diff
@@ -111,6 +126,9 @@ dependencies {
     // The version of react-native is set by the React Native Gradle Plugin
     implementation("com.facebook.react:react-android")
 
+    // Add MMKV dependency directly to make prefab available
+    implementation("com.tencent:mmkv-shared:2.2.2")
```
mrousavy, line 130: nice!

package/react-native-mmkv.podspec
```diff
@@ -19,20 +18,15 @@ Pod::Spec.new do |s|
...
   s.source_files = [
     # react-native-mmkv
     "ios/**/*.{h,m,mm}",
-    "cpp/**/*.{hpp,cpp,c,h}",
-    # MMKV/Core
-    "MMKV/Core/**/*.{h,cpp,hpp,S}",
+    "cpp/**/*.{hpp,cpp,c,h}"
   ]
 
+  s.dependency 'MMKVCore'
```
mrousavy, line 30: dope - so if the example app adds a dependency to `MMKV` it can use that from ObjC just fine?

package/scripts/folly_headers.sh
```diff
@@ -0,0 +1,160 @@
+#!/bin/bash
+
+set -e
+
+# Script to set up folly headers for React Native bridging
+# This is required because React Native's bridging headers include <folly/dynamic.h>
+# but React Native doesn't ship with the actual folly headers in node_modules
```
mrousavy, line 7: oh god.. do we absolutely need that? cc @hannojg

package/scripts/folly_headers.sh
```diff
@@ -0,0 +1,160 @@
...
+    
+    // Object/Array access (stubs)
+    dynamic& operator[](const std::string& key) { static dynamic d; return d; }
+    const dynamic& operator[](const std::string& key) const { static dynamic d; return d; }
+    dynamic& operator[](size_t index) { static dynamic d; return d; }
+    const dynamic& operator[](size_t index) const { static dynamic d; return d; }
+    
+    size_t size() const { return 0; }
+  };
+}
```
mrousavy, line 119: this is honestly just asking for trouble.. I bet $50 that this will break within a year

package/scripts/mmkv_headers.sh
```diff
@@ -0,0 +1,194 @@
+#!/bin/bash
+
+set -e
+
+# Script to set up MMKV headers for IDE support
+# This downloads or creates the necessary MMKV core headers without relying on the example app
```
mrousavy, line 6: This can't be correct. I believe @lingol added prefab support to the MMKV gradle package, which means it should ship C++ headers. See [their build.gradle prefab spec](https://github.com/Tencent/MMKV/blob/c74d8b886abd87132c507ec17865a7fb4feb9679/Android/MMKV/mmkv/build.gradle#L111-L115).

So we definitely don't want to download them, and downloading is also not the same version as the one we have locally - so the symbols won't match.

Are you doing this just for IDE support, or will those headers _actually_ be included?

package/scripts/mmkv_headers.sh
```diff
@@ -0,0 +1,194 @@
...
+        # Create minimal MMKV.h stub
+        cat > "$MMKV_HEADERS_DIR/MMKVCore/MMKV.h" << 'MMKV_STUB_EOF'
+#pragma once
+
+#include <string>
+#include <vector>
+#include <memory>
+
+// Minimal MMKV stub for IDE support
+// This provides basic interface for code completion
```
mrousavy, line 86: this will also break the second MMKV interface changes. not good

package/scripts/setup_clang_env.sh
```diff
@@ -0,0 +1,141 @@
...
+# Convert to absolute path
+BUILD_DIR="$(cd "$BUILD_DIR" && pwd)"
+
+# Set PKG_DIR to the packages/ directory
+PKG_DIR="$SCRIPT_DIR/../"
+
+# Convert to absolute path
+PKG_DIR="$(cd "$PKG_DIR" && pwd)"
+
+echo "Setting up React Native MMKV headers for IDE support..."
```
mrousavy, line 23: This also looks weird...

mrousavy (commented):
thanks for the PR brad! left a few notes

package/cpp/MmkvHostObject.cpp
```diff
@@ -221,12 +220,18 @@ jsi::Value MmkvHostObject::get(jsi::Runtime& runtime, const jsi::PropNameID& pro
           }
 
           std::string keyName = arguments[0].getString(runtime).utf8(runtime);
-          mmkv::MMBuffer buffer;
+          MMBufferType buffer;
+#ifdef __OBJC__
+          // iOS: Convert std::string to NSString* for MMKVCore pod compatibility
+          bool hasValue = instance->getBytes(@(keyName.c_str()), buffer);
```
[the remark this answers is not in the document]
boorad: If they could ship the C++ bits in the pod, we wouldn't need this.  The linker errors showed that while the MMKVCore pod provides the source files, it doesn't compile them into a linkable library for C++ usage. The pod is likely intended to be used from Objective-C/Swift code, not directly from C++.

boorad:
closing to target `main` branch in #847

PR #857, opened 2025-08-20
fix: support devices/simulators set to dark mode

boorad:
text will show up in Example app, if simulator/emulator is set to dark mode by default or during use.


mrousavy:
super thanks!

PR #858, opened 2025-08-20
feat: `android` 16KB page support

boorad:
I've seen other PRs use compiler flags.  Up to you how you want to do this...


mrousavy:
i dont think we need that as Codegen is building our CMakeLists project... i could be wrong

PR #869, opened 2025-08-28
perf: Use `mmkv-static` instead of `mmkv-shared`

mrousavy:


mrousavy:
Closing as stale: the Android/MMKV dependency setup has moved on and this branch is outdated.

Issue #861, opened 2025-08-20
Android build: Could not find a package configuration file provided by "mmkv" [4.0.0-beta.1, Expo 54, React Native 81]


coolsoftwaretyler:
Reproducer: https://github.com/coolsoftwaretyler/expo-54-react-native-mmkv-repro

If you pull down that repro, run:

1. `npm i`
2. `npx expo prebuild --clean`
3. `npm run android`

You'll get build errors like this:

```sh
> Task :app:configureCMakeDebug[arm64-v8a] FAILED
C/C++: CMake Error at /Users/tylerwilliams/blank54/node_modules/react-native-mmkv/android/CMakeLists.txt:24 (find_package):
C/C++:   Could not find a package configuration file provided by "mmkv" with any of
C/C++:   the following names:
C/C++:     mmkvConfig.cmake
C/C++:     mmkv-config.cmake
C/C++:   Add the installation prefix of "mmkv" to CMAKE_PREFIX_PATH or set
C/C++:   "mmkv_DIR" to a directory containing one of the above files.  If "mmkv"
C/C++:   provides a separate development package or SDK, be sure it has been
C/C++:   installed.

[Incubating] Problems report is available at: file:///Users/tylerwilliams/blank54/android/build/reports/problems/problems-report.html
```

I was originally experiencing the issues from https://github.com/mrousavy/react-native-mmkv/issues/849 when using Expo 54/RN 0.81.0/RN MMKV 3.3.0. I updated to 4.0.0-beta.1 and hit these new errors.

Let me know if you need any more details. Thanks for all the great work!

mrousavy:
The `mmkv` prefab should be contained in the `mmkv-shared` gradle plugin [added here](https://github.com/mrousavy/react-native-mmkv/blob/69679ea63f7b8855038366b3826cafff167b28b6/package/android/build.gradle#L89-L90).

Did you clean all your caches? Maybe re-clone the project and see if that changes anything. Or global gradle cache (`~/.gradle`). It builds for me, and here in CI.

mrousavy:
ok I can reproduce in your example;

````
> Task :app:configureCMakeDebug[arm64-v8a] FAILED
C/C++: CMake Error at /tmp/expo-54-react-native-mmkv-repro/node_modules/react-native-mmkv/android/CMakeLists.txt:24 (find_package):
C/C++:   Could not find a package configuration file provided by "mmkv" with any of
C/C++:   the following names:
[...]

PR #870, opened 2025-08-28
fix: include cpp folder in npm dist

coolsoftwaretyler:
This will fix [an Android build error with the new Nitro modules implementation of react-native-mmkv](https://github.com/mrousavy/react-native-mmkv/issues/861#issuecomment-3235206778).


mrousavy:
my bad, I fixed this on main before I saw your PR. But yes this is good thanks anyways

PR #871, opened 2025-08-29
Add MMKV class

riteshshukla04:
```
import { MMKV } from 'react-native-mmkv'

export const storage = new MMKV()
```

This lets us use this syntax. So hopefully migration from 3.x to 4.x.x should be easier


mrousavy:
This is not how the MMKV class should be implemented. I want the prototype / constructor to actually be the native instance.

riteshshukla04:
Lol Okay🫡

PR #872, opened 2025-08-29
feat: use jest not bun for tests, contributing docs

boorad:
- This PR ensures tests are run with jest, not bun
- Proper dependencies and configs were added to the monorepo root
- Upgraded deprecated function use with modern `@testing-library/react-native`
- `CONTRIBUTING.md` was updated for using `bun`


mrousavy:
this is good! thanks Brad

PR #880, opened 2025-09-04
Fix Expo plugin for android

leoboyerbx:
This PR fixes the Expo config plugin so that it generates the correct `gradle.properties`.

Included changes:
- Fixed the path in `app.plugin.js` to the correct plugin entry file
- Refactored the plugin to commonJS (with .cts extension) to avoid a bug with VSCode not recognising the plugin correctly in app.json
- Updated the property in-place instead of nuking the key and pushing a new one at the end of the file. This way, the generated file has the property next to its comment, like so:

```
# Use this property to specify which architecture you want to build.
# You can also override it from the CLI using
# ./gradlew <task> -PreactNativeArchitectures=x86_64
reactNativeArchitectures=arm64-v8a,x86_64
```

The generated `gradle.properties` is now correct, but it does not fix the Android build in an expo project. This seems to be a separate issue.

### DIFF
--- a/packages/react-native-mmkv/src/expo-plugin/withMMKV.cts
+++ b/packages/react-native-mmkv/src/expo-plugin/withMMKV.cts
@@ -1,11 +1,12 @@
-import type { ConfigPlugin } from '@expo/config-plugins'
-import { createRunOncePlugin, withGradleProperties } from '@expo/config-plugins'
+import type { ConfigPlugin, ExportedConfigWithProps } from '@expo/config-plugins'
+const { createRunOncePlugin, withGradleProperties } = require('@expo/config-plugins')
+import type { Properties } from '@expo/config-plugins/build/android'
 
 const pkg = require('../../package.json')
 
 const withMMKV: ConfigPlugin<{}> = (config) => {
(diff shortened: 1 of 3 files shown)


packages/react-native-mmkv/src/expo-plugin/withMMKV.cts
```diff
@@ -0,0 +1,31 @@
...
+    const index = cfg.modResults.findIndex(
+      (p) => p.type === 'property' && p.key === property.key
+    )
+    if (index !== -1) {
+      cfg.modResults[index] = property
+    }
+    // Else add it to the properties
+    else {
+      cfg.modResults.push(property)
+    }
```
mrousavy, line 26:
```suggestion
    if (index !== -1) {
      // Overwrite it
      cfg.modResults[index] = property
    } else {
      // Append it
      cfg.modResults.push(property)
    }
```

mrousavy:
Thanks, lgtm!

mrousavy:
Do you think you could add a GitHub action/CI test for this somehow so we can ensure the Expo Plugin works properly and won't break in the future? Thanks 🙏

PR #882, opened 2025-09-05
fix: change the return type of remove to bool from void

riteshshukla04:
The current remove function of `react-native-mmkv` returns void. 

We don't know if the key is removed or not. (Can happen in cases like instance is read only)
The `removeValueForKey` from MMKV core however returns a boolean value to handle such cases .


mrousavy:
LGTM! Thanks

PR #889, opened 2025-09-12
fix: platform extension conflict with mock file

timostroehlein:
A bit of a specific PR. In our project we use custom platform specific extensions. One of those extensions is a `.mock.ts` extensions which mocks all api endpoints. This package also contains a file with a platform specific `.mock.ts` extensions, but this shouldn't be one and because of that breaks that functionality in our app. The easiest fix would be to rename createMMKV.mock.ts to createMockMMKV.ts.


mrousavy:
That's fine I guess. Interesting issue

PR #887, opened 2025-09-08
feat: Use our MMKV fork to support 32-bit again

mrousavy:
This PR changes the Gradle dependency on MMKV from Tencent's original MMKV repo to our fork (https://github.com/margelo/MMKV) to add support for 32-bit Android devices again. Like TVs. 🥳 

This was the last blocker to make V4 stable!

PR #904, opened 2025-09-22
Update V4 upgrade guide with Expo plugin instructions

greg-schrammel:
Added instructions for using expo-build-properties plugin with Expo Continuous Native Generation.


mrousavy:
Turns out there is no reason to not support 32-bit. Everything builds properly for 32-bit architecture as well, so we forked MMKV to add 32-bit support back to MMKV, and now we don't need any extra configuration or expo plugins! :)

https://github.com/mrousavy/react-native-mmkv/pull/887

PR #908, opened 2025-09-23
Update MMKV usage in WRAPPER_JOTAI.md

robertherber:
Updated MMKV initialization and method names in Jotai integration. I guess the previous one was for a previous version of react-native-mmkv.


mrousavy:
This actually points to the V4 docs. I just pushed https://github.com/mrousavy/react-native-mmkv/commit/16f6cef63073794b677bc382117469354271a7c7, which makes the current README (V3) point to the V3 docs.

PR #932, opened 2025-10-20
chore: Add Jest Tests in Example App

riteshshukla04:
The current MMKV docs refers jest example which doesnt exist . Adding it here


mrousavy:
The tests were removed by @V3RON here: https://github.com/mrousavy/react-native-mmkv/pull/924 - do you think we should add them again?

riteshshukla04:
This is just to fix the code reference in docs.  `Testing with Jest`  section refers this. Either we can remove that section or add this back

PR #944, opened 2025-10-31
Mention mmkv being an in-memory store

ferretwithaberet:
Adds additional information about MMKV being an in-memory store backed by `mmap` to avoid confusion. This gives more context for people that do not know what `mmap` or `MMKV` are or how they work.
I, personally, made the mistake of persisting tanstack query's cache with `maxAge: Infinity` to `react-native-mmkv`, without ever realizing that it is an in-memory store and that I am filling the RAM of my users' devices.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@
 <br/>
 
 
-* **MMKV** is an efficient, small, in-memory synced with files (backed by mmap) mobile key-value storage framework developed by WeChat. See [Tencent/MMKV](https://github.com/Tencent/MMKV) for more information
+* **MMKV** is an efficient, small, **in-memory** mobile key-value storage framework developed by WeChat. Data is stored using `mmap` which syncs it with files for persistence. See [Tencent/MMKV](https://github.com/Tencent/MMKV) for more information
 * **react-native-mmkv** is a library that allows you to easily use **MMKV** inside your React Native app through fast and direct JS bindings to the native C++ library.
 
 ## Features
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@
 <br/>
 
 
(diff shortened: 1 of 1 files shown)


XChikuX:
I agree. I didn't know the fetches were happening from RAM. Especially if all data is stored in ram without LRU cache. (not sure if this is how it works)

However, I think your explanation complicates understanding. @ferretwithaberet.

I'd prefer you keep the line:

* **MMKV** is an efficient, small mobile key-value storage framework developed by WeChat. See [Tencent/MMKV](https://github.com/Tencent/MMKV) for more information

and add

> Stored data is loaded onto memory using mmap and synced with the filesystem for persistence.

As an additional line. Please correct me if I've made any mistakes in understanding.

LovesWorking:
Yea I did the same thing a few years ago!

ferretwithaberet:
> I agree. I didn't know the fetches were happening from RAM. Especially if all data is stored in ram without LRU cache. (not sure if this is how it works)
> 
> However, I think your explanation complicates understanding. @ferretwithaberet.
> 
> I'd prefer you keep the line:

It's not meant to be the clearest description. I meant for it to bring the "in-memory" word in the user's face, this should raise some question marks and they will dig deeper (by clicking on the MMKV link hopefully).

>     * **MMKV** is an efficient, small mobile key-value storage framework developed by WeChat. See [Tencent/MMKV](https://github.com/Tencent/MMKV) for more information
> 
> 
> and add
> 
> > Stored data is loaded onto memory using mmap and synced with the filesystem for persistence.

As mentioned above, I'd rather keep the description small and concise with few important keywords. Users will likely tend to skip over long blocks of text so they might miss the information. I'd rather make the "in-memory" word bold, so they instantly see it. I also wanted to keep the "synced with files" part and the mmap mention so users get an ideea of how it works, I guess the mmap part could be omitted and mentioned somewhere else, but the "synced with files" part should stay, maybe just reworded.

PR #950, opened 2025-11-05
Replace Flipper with Rozenite for MMKV debugging

retyui:
Updated integration section to reflect Rozenite usage instead of Flipper.


retyui:
could you please review this PR @mrousavy ?

mrousavy:
It's been less than 2 days, have some patience.

Issue #956, opened 2025-11-10
Removing a key on web always return false


loiclouvet:
### Description

`MMKV.remove` on the **Web** always returns false and never triggers listeners, even if the key exists.
The root cause is that `Storage.removeItem` returns `void`, so the current implementation interprets the result as falsy and assumes the delete failed. Listeners only fire when remove reports true, so they never run.

### Expected Behavior

`MMKV.remove` should return true  and fire event when a key existed.

### Actual Behavior

The key is deleted, but remove returns false and no listener runs, making client code believe nothing changed.

### Environment

```
react-native-mmkv: v4.0.0
Chrome Version 142.0.7444.60
```

### Suggested patch:
```
diff --git a/lib/createMMKV/createMMKV.web.js b/lib/createMMKV/createMMKV.web.js
index 82be040ecc28fed9168d42db2641c6e1e32c35f1..1db1c97522c7223a28475858dc04dd930e981fe5 100644
--- a/lib/createMMKV/createMMKV.web.js
+++ b/lib/createMMKV/createMMKV.web.js
@@ -69,10 +69,12 @@ export function createMMKV(config = { id: 'mmkv.default' }) {
             }
         },
         remove: (key) => {
-            const wasRemoved = storage().removeItem(prefixedKey(key)) ?? false;
-            if (wasRemoved)
-                callListeners(key);
-            return wasRemoved;
+            const keyWithPrefix = prefixedKey(key)
+            const existed = storage().getItem(keyWithPrefix) != null
+            storage().removeItem(keyWithPrefix)
+            if (existed)
+                callListeners(key)
+            return existed
         },
         set: (key, value) => {
             if (key === '')
```

mrousavy:
Thanks - can you please create a PR for the patch instead of posting it here?

PR #957, opened 2025-11-13
fix: Fix incorrect return of  on Web

riteshshukla04:
Should fix https://github.com/mrousavy/react-native-mmkv/issues/956


mrousavy:
thanks!

PR #959, opened 2025-11-19
Remove z and c++ libraries from podspec

riteshshukla04:
This is not needed since we dont build it from source anymore. In Most cases it will be removed automatically though


mrousavy:
Ah - didn't know, thanks!

PR #960, opened 2025-11-21
docs: update react-query integration to use async persister

fobos531:
https://tanstack.com/query/latest/docs/framework/react/plugins/createSyncStoragePersister

CleanShot 2025-11-21 at 19 40 55@2x


mrousavy:
thanks!

PR #964, opened 2025-11-28
chore: make clearAll bit faster

riteshshukla04:
As discussed on discord, This PR optimises clearAll by skipping listener notification when no listeners exist


mrousavy:
I dont think this is worth the added complexity, but thanks!

PR #983, opened 2026-01-09
fix: equals method override

agudeluca:
# Fix: Remove `override` Keyword from `equals` Method

## Summary

This fix removes the `override` keyword from the `equals` method in `HybridMMKVPlatformContextSpecSwift.hpp` to resolve a compilation error.

## Problem

The `equals` method in `HybridMMKVPlatformContextSpecSwift` was declared with the `override` keyword:

```cpp
bool equals(const std::shared_ptr<HybridObject>& other) override {
  if (auto otherCast = std::dynamic_pointer_cast<HybridMMKVPlatformContextSpecSwift>(other)) {
    return _swiftPart.equals(otherCast->_swiftPart);
  }
  return false;
}
```

However, the base class `HybridMMKVPlatformContextSpec` does not declare `equals` as a virtual function. The `override` keyword in C++ requires that the method being overridden is actually declared as `virtual` in a base class. Since this condition is not met, the compiler would generate an error.

## Solution

Removed the `override` keyword from the method declaration:

```cpp
bool equals(const std::shared_ptr<HybridObject>& other) {
  if (auto otherCast = std::dynamic_pointer_cast<HybridMMKVPlatformContextSpecSwift>(other)) {
    return _swiftPart.equals(otherCast->_swiftPart);
  }
  return false;
}
```

### Class Hierarchy

- `HybridMMKVPlatformContextSpecSwift` inherits from `HybridMMKVPlatformContextSpec`
- `HybridMMKVPlatformContextSpec` inherits from `HybridObject`
- The `equals` method is not declared as `virtual` in `HybridMMKVPlatformContextSpec`

### Comparison with Android Implementation

The Android equivalent (`JHybridMMKVPlatformContextSpec`) correctly uses `override` because it inherits from `JHybridObject`, which likely declares `equals` as a virtual function. The iOS Swift implementation has a different inheritance structure, so it doesn't need (and shouldn't have) the `override` keyword.

## Files Changed

- `packages/react-native-mmkv/nitrogen/generated/ios/c++/HybridMMKVPlatformContextSpecSwift.hpp`

## Impact

- **Before**: Compilation error due to incorrect use of `override` keyword
- **After**: Code compiles successfully, method behavior remains unchanged

## Related

- Pull Request: #983
- File: `nitrogen/generated/ios/c++/HybridMMKVPlatformContextSpecSwift.hpp`
- Line: 51

### DIFF
--- a/packages/react-native-mmkv/nitrogen/generated/ios/c++/HybridMMKVPlatformContextSpecSwift.hpp
+++ b/packages/react-native-mmkv/nitrogen/generated/ios/c++/HybridMMKVPlatformContextSpecSwift.hpp
@@ -48,7 +48,8 @@ namespace margelo::nitro::mmkv {
     inline size_t getExternalMemorySize() noexcept override {
       return _swiftPart.getMemorySize();
     }
-    bool equals(const std::shared_ptr<HybridObject>& other) override {
+
+    bool equals(const std::shared_ptr<HybridObject>& other) {
       if (auto otherCast = std::dynamic_pointer_cast<HybridMMKVPlatformContextSpecSwift>(other)) {
         return _swiftPart.equals(otherCast->_swiftPart);
       }


mrousavy:
The base class does define `equals` as a virtual function.

You just need to update your react-native-nitro-modules version.

Also, editing generated files is never acceptable.

PR #995, opened 2026-02-10
feat: Make default MMKV log level configurable at build time

tbrushwyler:
Allow library consumers to override the default MMKV core log level via Gradle properties (Android) or Podfile variables (iOS).

### DIFF
--- a/packages/react-native-mmkv/cpp/MMKVTypes.hpp
+++ b/packages/react-native-mmkv/cpp/MMKVTypes.hpp
@@ -51,12 +51,12 @@ using namespace mmkv;
 
 // Default MMKV log level if not configured at build time
 #ifndef MMKV_LOG_LEVEL
-  #ifdef NITRO_DEBUG
-    #define MMKV_LOG_LEVEL 0
-  #else
-    #define MMKV_LOG_LEVEL 3
-  #endif
+#ifdef NITRO_DEBUG
+#define MMKV_LOG_LEVEL 0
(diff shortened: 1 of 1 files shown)


mrousavy (changes requested):
Thank you for this PR - this is great stuff!

I just have one change - can we not make the `MMKV_LOG_LEVEL` definition optional and instead always define it, but if it's not defined we use a default value (`0` in debug, `3` in release)?

That way we can eliminate some of the conditionals here - always good to remove conditionals for predictability.

tbrushwyler:
> Thank you for this PR - this is great stuff!
> 
> I just have one change - can we not make the `MMKV_LOG_LEVEL` definition optional and instead always define it, but if it's not defined we use a default value (`0` in debug, `3` in release)?
> 
> That way we can eliminate some of the conditionals here - always good to remove conditionals for predictability.

Good idea - I've moved this to `MMKVTypes.hpp` so that it's discoverable and falls back to the existing value, controlled by `NITRO_DEBUG`.

tbrushwyler:
@mrousavy could you re-trigger the Build iOS job? It looks like it failed for an unrelated reason:

https://github.com/mrousavy/react-native-mmkv/actions/runs/21909536593/job/63378522905?pr=995
Screenshot 2026-02-18 at 8 40 00 AM

mrousavy:
Awesome, thanks! LGTM

Issue #1006, opened 2026-02-22
useMMKV hooks return stale value when used with React <Activity/>


KijongHan:
minimum repro setup here with source code changes in the example folder under `App.tsx`:
https://github.com/KijongHan/react-native-mmkv/tree/reproduce/useMMKV-hooks-returning-stale-values

Two test components are set up so that TestB is under an `Activity` component whose visibility mode toggles on/off depending on a react state. TestB subscribes using useMMKV hook. TestA is under an Activity that is always visible and also subscribes to the same hook and key subscribed for TestB

when you update the key value from the parent while TestB is not visible, TestA listens to the updated changes as expected. When you toggle visibility for TestB, it reads and listens to a stale value for the mmkv key

mrousavy:
I think this is now fixed with https://github.com/mrousavy/react-native-mmkv/pull/1008 - can you confirm?

KijongHan:
awesome thanks @mrousavy! yeah `useSyncExternalStore` implementation fixes this issue. Any ideas please as to the timeline for a new release?

PR #1007, opened 2026-02-22
fix: trigger bump when useEffect is mounted for the first time to retrieve latest value and prevents stale values being returned

KijongHan:
I had a stab at resolving [this issue](https://github.com/mrousavy/react-native-mmkv/issues/1006) and from local testing it no longer listens to stale values when components are remounted under `Activity`


mrousavy:
Thanks for the PR, but I think this is outdated and needs to go. We can replace this with `useSyncExternalStore` instead, which should be much more reliable; I drafted a PR via Codex: https://github.com/mrousavy/react-native-mmkv/pulls

PR #1032, opened 2026-04-02
feat: Add support for Harness web

riteshshukla04:
feat: Add support for Harness web

### DIFF
--- a/.github/workflows/build-ios.yml
+++ b/.github/workflows/build-ios.yml
@@ -30,8 +30,6 @@ on:
 
 env:
   USE_CCACHE: 1
-  # Must match the runner's architecture (macOS-15 runners are arm64)
-  ARCH: arm64
 
 jobs:
   build:
@@ -77,6 +75,6 @@ jobs:
           -scheme MmkvExample \
           -sdk iphonesimulator \
           -configuration Debug \
-          -destination 'generic/platform=iOS Simulator,arch=${{ env.ARCH }}' \
+          -destination 'generic/platform=iOS Simulator' \
           build \
           CODE_SIGNING_ALLOWED=NO"
--- a/.github/workflows/build-ios.yml
+++ b/.github/workflows/build-ios.yml
@@ -54,8 +54,8 @@ jobs:
           bundler-cache: true
           working-directory: example
 
-      - name: Select Xcode 16.4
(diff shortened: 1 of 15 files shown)


example/__tests__/MMKV.harness.ts
```diff
@@ -399,7 +404,7 @@ describe('MMKV Configuration & Multiple Instances', () => {
       instances.forEach(instance => instance.clearAll());
     });
 
-    it('should import other keys properly', () => {
+    itNativeOnly('should import other keys properly', () => {
```
mrousavy: Please revert all of this *nativeOnly stuff.

If something is intentionally not supported on web, guard the whole test/it/describe behind a if(IS_NATIVE) flag.

This is impossible to read for the future once we add support for such things.

Also, for things that _should_ be shpported on the web (like ArrayBuffers), remove the restriction. If it fails now we fix it later.

mrousavy (changes requested):
Overall this is a good direction, but I have two blockers before merging:

1. My comment about nativeOnly tests
2. The patch. With a patch, this can't go into main. Create an upstream PR in Harness / RN for this, talk to Szymon Chmal or Hanno about this

example/patches/harness-web-support.patch
```diff
@@ -0,0 +1,141 @@
...
++import { Platform } from 'react-native';
++
+ export const getDevServerUrl = () => {
++    if (Platform.OS === 'web' && typeof window !== 'undefined') {
++        return window.location.origin;
++    }
++    const getDevServer = require('react-native/Libraries/Core/Devtools/getDevServer');
+     const devServer = getDevServer();
+     return devServer.url;
+ };
```
mrousavy: Did you create Harness PRs for all of this?

.github/workflows/build-ios.yml
```diff
@@ -1,5 +1,6 @@
 name: Build iOS
 
+
```
mrousavy:
```suggestion
```

.github/workflows/build-ios.yml
```diff
@@ -30,8 +31,7 @@ on:
 
 env:
   USE_CCACHE: 1
-  # Must match the runner's architecture (macOS-15 runners are arm64)
-  ARCH: arm64
+
```
mrousavy:
```suggestion
```

PR #1058, opened 2026-05-27
chore: Remove unused Expo config plugin dependency

mrousavy:
## Summary
- Remove unused `@expo/config-plugins` from the library devDependencies.
- Refresh `bun.lock` to drop its Expo config-plugin transitive graph.

## Notes
- The package manifest lists `app.plugin.js` in `files`, but no such file exists and there are no imports/usages of `@expo/config-plugins` in the repo.
- Latest `@expo/config-plugins` is `56.0.8`, but upgrading it would keep a dead dependency around.

## Validation
- `rg -n "expo|config-plugins|withPlugins|ConfigPlugin|app.plugin|plugins" . -g '!node_modules' -g '!example/ios/Pods'`
- `bun install`
- `bun typecheck`
- `bun lint-ci`
- `bun run test`
- `git diff --check`

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -61,7 +61,6 @@
       "name": "react-native-mmkv",
       "version": "4.3.1",
       "devDependencies": {
-        "@expo/config-plugins": "^10.1.2",
         "@react-native/eslint-config": "0.85.3",
         "@react-native/jest-preset": "0.85.3",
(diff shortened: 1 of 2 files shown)


mrousavy:
Superseded by #1063, which folds the package upgrades into one PR.

